PR #2, opened 2020-10-12
chore: Initialize project

jkayani:
- Makefile to do the CI chores

- Cobra basic scaffolding for our CLI: https://github.com/spf13/cobra. I did it manually rather than via the code generator just to learn

- GH actions pipeline

Should we add something to help standup ArgoCD locally easily? We'll need that at some point to do testing of this plugin


.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,16 @@
+name: Basic pipeline
+on: push
```
werne2j, line 2: Don't forget to add `pull requests` for when to run the pipeline

jkayani:
Do we want to get advanced and configure merging to master to create GH releases?

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
```
werne2j, line 9: We will also want to run `go vet` as well as `go test`

jkayani: Left out `go test` for now since we don't have any tests.

Makefile
```diff
@@ -0,0 +1,24 @@
...
+VERSION=0.1
+OS_ARCH=darwin_amd64
+
+default: build
+
+format: 
+	go fmt github.com/IBM/argocd-vault-plugin/...
+
+quality:
+	go vet
```
werne2j, line 11: I think you need to do `go vet ./...`

jkayani: Yeah I had a suspicion but it didn't seem to do anything either way, thanks for telling me

Makefile
```diff
@@ -0,0 +1,24 @@
+BINARY=argocd-vault-plugin
+VERSION=0.1
+OS_ARCH=darwin_amd64
+
+default: build
+
+format: 
+	go fmt github.com/IBM/argocd-vault-plugin/...
```
werne2j, line 8: I don't think we need this, unless we are stopping the build on code that is not in the correct format https://github.com/spf13/cobra/blob/master/Makefile#L16

jkayani: You're right, I put that in there thinking it was a linter not a formatter

jkayani: Do we want to fail on bad format?

werne2j: I dont think we need that now. Most IDEs have auto formatting anyways.

jkayani: Note for Josh: setup auto format in VS Code

werne2j (approved):

PR #6, opened 2020-10-26
feat: Add generate command, a sample test and some docs outline

werne2j:


jkayani (approved):
Looks good, but can we also add an invocation of the tests to CI? https://github.com/IBM/argocd-vault-plugin/blob/main/Makefile#L7

PR #15, opened 2020-11-16
feat: support Deployments and Secrets

jkayani:
This PR adds support for templated Deployment and Secret YAML files. Please review for:

- Structure - if the code structure is approved, I will use this to implement the other resources (ConfigMap, CronJob, Job, etc)

- Proper error handling/messaging

- Soundness of logic

Design decisions:

- I chose to use a `Template` interface all the different templates could implement so that from the CLI, they could be used the same, while giving us flexibility to implement each template's find/replace logic separately if needed. This is useful for Deployments vs Secrets, since the latter requires `[]byte` replacement values rather than `interface{}`. 

- I embedded a `Resource` type in each template since each template will need the fields of that `Resource`. It also helped me write sufficiently generic find/replace logic

There aren't any tests for this feature. Once this overall structure is approved I will proceed with adding tests and implementing the other resources. 

It works:
```
Joshuas-MBP-2:argocd-vault-plugin jkayani$ ./argocd-vault-plugin generate fixtures/nonempty/

apiVersion: apps/v1
kind: Deployment
metadata:
  creationTimestamp: null
  labels:
    app: inspector
    version: alpha
  name: inspector
  namespace: default
spec:
  minReadySeconds: 5
  replicas: 3
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: inspector
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: inspector
      name: inspector
    spec:
      containers:
      - image: foo.com/inspector:123
        imagePullPolicy: Always
        name: inspector
        resources: {}
status: {}
---
apiVersion: v1
data:
  SECRET_NUM: NQ==
  SECRET_VAR: Zm9v
kind: Secret
metadata:
  creationTimestamp: null
  name: inspector
  namespace: default
type: Opaque
---
```


cmd/generate.go
```diff
@@ -1,21 +1,118 @@
...
+				return errors.New("<path> argument required to generate manifests")
+			}
 			return nil
 		},
 	}
 
 	return command
 }
+
+func listYamlFiles(root string) []string {
```
werne2j, line 75: I would say we should move these methods to a util file as well

cmd/generate.go
```diff
@@ -1,21 +1,118 @@
...
+	var files []string
+
+	err := filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if filepath.Ext(path) == ".yaml" {
+			files = append(files, path)
+		}
+		return nil
+	})
+	if err != nil {
+		panic(err)
```
werne2j, line 85: No `panic`, we want to make sure to return the error and log it  We always want to return a meaningful error

cmd/generate.go
```diff
@@ -1,21 +1,118 @@
...
+			}
+			manifests, errs := readFilesAsManifests(files)
+			if len(errs) != 0 {
+
+				// TODO: handle multiple errors nicely
+				return fmt.Errorf("could not read YAML files: %s", errs)
+			}
+
+			var resource kube.Template
+			for _, manifest := range manifests {
```
werne2j, line 37: I would say lets pull this switch out and have a method that returns a `Template` interface

werne2j (changes requested):

cmd/generate.go
jkayani: Agreed about meaningful errors, but I guess I wonder what the purpose of `panic` actually is then? I thought it was for unrecoverable situations. Like why ever `panic` if you can just return `error`'s all the way back up?

jkayani: FYI there is still a `panic(error)` in `pkg/kube/util`'s `replaceInner` in the cases where we cannot convert to an inner YAML node to `map[string]interface{}`. Is that OK? I think I put that there b/c that was unexpected situation and thought panic was appropriate

werne2j: We will just need to see how it handles within Argo but it’s probably fine for now

werne2j (approved):

Issue #32, opened 2020-12-10
Use of config file


werne2j:
We want to allow the use of a config file instead of just environment variables for things like vault path, vault path, auth type, etc. May want to look into https://github.com/spf13/viper for accomplishing this.

jkayani:
Viper looks like a good choice, since it will abstract away where the config comes from (env variable or config file). 

It seems that normal use involves reading config files from disk; however, in our usage, we want to read YAMLs embedded in Kubernetes secrets right (since the full configuration will contain credentials)? In that case, should we expect users to mount their secret in a certain file path in their argocd-repo-server, and read it that way? Or maybe just have the user mount the secret as file somewhere in the repo server, and then have them pass the path to the configuration as a command line argument when invoking the plugin?

I'm mentioning mounting secrets as files so that we don't have to ask the ArgoCD server's Kubernetes API for that secret's value (and therefore  deal with permissions and a service account), and so that we easily allow for config-as-a-file for non-ArgoCD users.

werne2j:
I think we should allow passing a config file to the cli via a path for sure.

However, I’m not sure we wouldn’t be able to read a secret directly from kubernetes though as we are inside the Argo repo server pod which should already have the permissions we would need.

jkayani:
You're right - I keep forgetting this is a binary running in a pod, not another pod itself. Question still remains though, is it sufficient to accept a config file via CLI, use Viper to combine that info with any environment variables to create a configuration we use in the app, and expect people who want to store their config files in Kubernetes to use volume mounts? 

Or should we accept both a config file argument or the name of a Kubernetes secret in the current namespace, and in case of the latter, call the API to get the secret data?

werne2j:
> Or should we accept both a config file argument or the name of a Kubernetes secret in the current namespace, and in case of the latter, call the API to get the secret data?

This

jkayani:
Last question - now that we have Viper, we can map multiple env variables to a single config parameter easily. So this would allow for us to read a user's IBM Cloud API key from multiple possible env variables, with a precedence - should we use this to allow things like `AVP_IBM_API_KEY` as well as something more generic and accepted by other tools, like `BX_API_KEY`?

werne2j:
I guess i will leave that up to you. I decided on use the `AVP` prefix so we could have things isolated and scoped. But if you think that is a good idea, go for it.

jkayani:
[...]

PR #33, opened 2021-01-06
feat: Read configuration from files, Secrets, and env variables

jkayani:
### Description

Previously we read configuration for Vault from env variables. Now we can read from files on disk, and from Kubernetes Secrets directly. 

Accomplished by adding 2 new CLI flags:

- `-c` to specify the path to a config file
- `-s` to specify the name of a secret in `argocd` namespace of the ArgoCD cluster

A config file can be in any format Viper accepts (JSON, YAML, etc). The keys must be the ones shown in this YAML example:
```
VaultType: vault
Address: http://a-vault/
PathPrefix: my-app
AuthType: github
GithubToken: token
RoleID: blah
SecretID: big-blah
IBMCloudAPIKey: biggest-blah
```

A Secret just has to have the above keys:
```
apiVersion: v1
data:
  Address: aHR0cHM6Ly92YXVsdC5ib2Iv
  AuthType: Z2l0aHVi
  GithubToken: Zm9v
  PathPrefix: YmxhaAo=
  VaultType: dmF1bHQ=
kind: Secret
metadata:
  name: bob
  namespace: argocd
type: Opaque
```

Set env variables take precedence over config files and k8s Secrets via Viper. The env variable names we accept should be the same exact ones accepted before. Note that reading k8s Secrets from the plugin requires the `argocd-repo-server` to have a service account token mounted in the standard location `/var/run/secrets/kubernetes.io/`, which by default it doesn't do for some reason.

Also added error messages when a required configuration parameter is left empty; i.e, `GithubToken` for `github` authentication into a `vault` Vault. 

**Fixes:** #32 

### Checklist
- [ ] Tests for the changes have been

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] New tests
- [ ] Other (please describe)


pkg/vault/config.go
```diff
@@ -16,18 +16,31 @@ type Config struct {
 	Type       VaultType
 }
 
+func setupViper() {
+	viper.BindEnv("VaultType", "AVP_TYPE")
```
werne2j, line 20: We should be able to use the `SetEnvPrefix` to simplify this.
https://github.com/spf13/viper#working-with-environment-variables

pkg/client/kubernetes.go
```diff
@@ -0,0 +1,42 @@
...
+	"k8s.io/client-go/rest"
+	k8yaml "sigs.k8s.io/yaml"
+)
+
+type Client struct {
+	client *kubernetes.Clientset
+}
+
+func NewClient() (*Client, error) {
+	config, err := rest.InClusterConfig()
```
werne2j, line 15: Can we have a nice error in the situation where someone tries to do this while not in kubernetes

cmd/generate.go
```diff
@@ -71,5 +96,7 @@ func NewGenerateCommand() *cobra.Command {
 		},
 	}
 
+	command.Flags().StringVarP(&configPath, "config-path", "c", "", "path to a configuration file (YAML, JSON, envfile) to use")
+	command.Flags().StringVarP(&secretName, "secret-name", "s", "", "name of a Kubernetes Secret containing Vault configuration data in the argocd namespace of your ArgoCD host")
```
werne2j, line 100: Can you add something to make it very clear that `secret-name` will only work if in ArgoCD?

werne2j (commented):
Overall looks pretty good. Just a few comments.

pkg/vault/config.go
jkayani: I looked at that but:

> If the ENV variable name is not provided, then Viper will automatically assume that the ENV variable matches the following format: prefix + "_" + the key name in ALL CAPS. When you explicitly provide the ENV variable name (the second parameter), it does not automatically add the prefix.

We'd have to either rename the Viper keys from `RoleID` to `Role_ID` to match the env var name `AVP_ROLE_ID` or rename the env variables. We can go either way I just didn't do them.

pkg/client/kubernetes.go
```diff
@@ -0,0 +1,42 @@
+package client
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	k8yaml "sigs.k8s.io/yaml"
+)
+
+type Client struct {
```
jkayani, line 12: Is there a better name for this? Maybe package `localk8s`?

pkg/vault/config.go
jkayani: Depending on how we answer this, we may or may not need to add docs that say what the accepted keys in a config file /Secret are

pkg/client/kubernetes.go
jkayani: Added

cmd/generate.go
werne2j: I was just looking for something like `(Only available when used in ArgoCD)` to make it very obvious. Someone might read this `name of a Kubernetes Secret containing Vault configuration data in the argocd namespace of your ArgoCD host` and not really understand.

pkg/vault/config.go
werne2j: > rename the Viper keys from RoleID to Role_ID
I am fine with this

jkayani:
Seems this code is failing the `go vet` check with:

```
# github.com/hashicorp/vault/vault/quotas
../../../go/pkg/mod/github.com/hashicorp/vault@v1.6.1/vault/quotas/quotas_rate_limit.go:280:33: assignment mismatch: 4 variables but rlq.store.Take returns 5 values
../../../go/pkg/mod/github.com/hashicorp/vault@v1.6.1/vault/quotas/quotas_rate_limit.go:280:50: not enough arguments in call to rlq.store.Take
	have (string)
	want (context.Context, string)
../../../go/pkg/mod/github.com/hashicorp/vault@v1.6.1/vault/quotas/quotas_rate_limit.go:305:25: not enough arguments in call to rlq.store.Close
	have ()
	want (context.Context)
```

I'm not really sure what to do here. I tried setting the Vault package to 1.6.1 since it was 1.6.0 before and one of my previous commits downgraded it to 1.4.something to fix this.

pkg/client/kubernetes.go
werne2j: Can be renamed, not a fan of `LocalK8s` can be `KubernetesClient` or something

werne2j:
Maybe we should not run vet on the dependencies

README.md
```diff
@@ -42,17 +42,58 @@ data:
...
+The plugin requires some configuration to connect to Vault. The parameters are:
 
-| Name            | Description                | Supported            |
+| Name            | Description                | Notes                |
 | --------------- | -------------------------- | -------------------- |
-| AVP_VAULT_ADDR  | Address of your Vault      | N/A                  |
-| AVP_PATH_PREFIX | Prefix of the vault path to look for the secrets | N/A       |
-| AVP_TYPE        | The type of Vault backend  | `vault` and `secretmanager` |
-| AVP_AUTH_TYPE   | The type of authentication | vault: `approle, github`   secretmanager: `iam` |
+| VAULT_ADDR     | Address of your Vault      | N/A                  |
```
werne2j, line 50: These still have to have the prefix, right?

jkayani: Yeah only when specified as env variables

werne2j: Oh so its not needed when a config or secret, got it

werne2j (approved):
Looks good

Issue #31, opened 2020-12-10
Use token if still valid


werne2j:
Right now we get a new token everytime the tool is called. We should store the token and check if its still valid and using that if it is still valid before trying to get a new token

werne2j:
Will have to write to a file, not able to store the value as a env var. Wondering if we can use the same config file for the token and the required params. Or even if this work is truly necessary.

jkayani:
Using this https://github.com/spf13/viper#writing-config-files along with `viper.Set("VAULT_TOKEN", token)` could help

PR #35, opened 2021-01-07
feat: Read/Write vault token to/from file

werne2j:
### Description
Write vault token to a file that can then be read on subsequent runs. Will check the tokens expire date and if expired will re authenticate and write new token.

**Fixes:** #31 

### Checklist
- [x] Tests for the changes have been

### Type of Change

- [ ] Bugfix
- [x] Feature
- [x] New tests
- [ ] Other (please describe)

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -1,11 +1,9 @@
 package cmd
 
 import (
-	"bytes"
 	"fmt"
 
 	"github.com/IBM/argocd-vault-plugin/pkg/kube"
-	"github.com/IBM/argocd-vault-plugin/pkg/kubernetesclient"
 	"github.com/IBM/argocd-vault-plugin/pkg/vault"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -41,12 +39,13 @@ func NewGenerateCommand() *cobra.Command {
 				return fmt.Errorf("could not read YAML files: %s", errs)
 			}
 
-			err = setConfig(secretName, configPath)
+			viper := viper.New()
+			err = setConfig(secretName, configPath, viper)
 			if err != nil {
 				return err
 			}
 			//
-			vaultConfig, err := vault.NewConfig()
+			vaultConfig, err := vault.NewConfig(viper)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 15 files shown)


werne2j:
Still need to write tests but wanted to get eyes on it

cmd/generate.go
```diff
@@ -41,40 +41,26 @@ func NewGenerateCommand() *cobra.Command {
...
-				viper.SetConfigType("yaml")
-				viper.ReadConfig(bytes.NewBuffer(yaml))
-			}
-			if configPath != "" {
-				viper.SetConfigFile(configPath)
-				err := viper.ReadInConfig()
-				if err != nil {
-					return err
-				}
+			err = setConfig(secretName, configPath)
```
jkayani, line 44: Ha, thanks

pkg/vault/util.go
```diff
@@ -0,0 +1,79 @@
...
+		byteValue, _ := ioutil.ReadAll(jsonFile)
+
+		var result map[string]interface{}
+		json.Unmarshal([]byte(byteValue), &result)
+
+		vaultConfig.VaultAPIClient.SetToken(result["vault_token"].(string))
+		tokenself, _ := vaultConfig.VaultAPIClient.Auth().Token().LookupSelf()
+		expireTime := tokenself.Data["expire_time"]
+		today := time.Now()
+		myDate, _ := time.Parse("2021-01-08T08:21:27.195176037Z", expireTime.(string))
```
jkayani, line 40: What is the specific date string used for here?

pkg/vault/util.go
```diff
@@ -0,0 +1,79 @@
...
+		return err
+	}
+
+	path := filepath.Join(home, ".avp")
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		os.Mkdir(path, 0755)
+	}
+
+	viper.Set("VAULT_TOKEN", token)
+	err = viper.WriteConfigAs(filepath.Join(path, "config.json"))
```
jkayani, line 71: Interesting, we use Viper to store and write the token to a file but then we don't use Viper to read it back, I guess so that we don't change any other Viper parameters?

pkg/vault/util.go
werne2j: `Parse parses a formatted string and returns the time value it represents. The layout defines the format by showing how the reference time, defined to be` - https://golang.org/pkg/time/#Parse

pkg/vault/util.go
werne2j: Yeah, i was thinking reading that file in would override. I could change to not use viper to write the file for the sake of continuity

jkayani: I guess that would make sense since we're not reading with Viper

jkayani: You could also setup another instance of Viper to make reading/writing easier

jkayani:
This code looks fine to me. Only suggestion is the one in the Viper thread

pkg/vault/util.go
werne2j: > You could also setup another instance of Viper to make reading/writing easier

Didn't see a way to do this (Or at least i didn't comprehend how to do that)

werne2j: https://github.com/spf13/viper#working-with-multiple-vipers i guess i missed this

cmd/generate.go
```diff
@@ -41,40 +39,27 @@ func NewGenerateCommand() *cobra.Command {
...
-					return err
-				}
+			viper := viper.New()
+			err = setConfig(secretName, configPath, viper)
+			if err != nil {
+				return err
 			}
-
-			config, err := vault.NewConfig()
+			//
```
jkayani, line 47: Nit

jkayani (approved):
Looks good!

Issue #54, opened 2021-02-08
Option to ignore if no secret for type found


jamesmmchugh:
**Is your feature request related to a problem? Please describe.**
The plugin tries to find secrets by appending the resource type to the path_perfix. If, however, you have no secrerts for a particular resource/resource type then you get errors like the following (using a namespace as the example):

```
Error making API request.

URL: GET http://<vault_server>/v1/namespace
Code: 403. Errors:

* 1 error occurred:
	* permission denied
```

**Describe the solution you'd like**
For the plugin to ignore resources that have no secrets templated and simply output them as they aer

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

jkayani:
Hi,

To confirm, you believe that in the scenario where you have a YAML like:

```
kind: ConfigMap
data:
 hello: world
```

in a folder that you are processing via the plugin, and there is no Vault secret at `${PATH_PREFIX}/configmap` the plugin should simply output the file as-is, right? If so, that makes sense to me. 

But what about this scenario? A YAML like:

```
kind: ConfigMap
data:
 hello: <world>
```

where someone forgets to create the Vault secret at `${PATH_PREFIX}/configmap` - then I guess you'd want to see an explicit 4xx where the plugin tried to replace `<world>`, instead of seeing an error trying to apply a YAML with `<world>` literally in it. Right?

If so, I think the desired behavior altogether is:

- If a given `Template` (YAML) has placeholders, and Vault returns an error trying to retrieve data for it, return that error so the user has a hint on how to fix it (create an entry in Vault for that resource or remove the `<placeholder>'s`)

- If a given `Template` has _no_ placeholders, emit the YAML as-is

jkayani:
Sorry if the above seems long-winded, I'm trying to keep in mind the exact logic the code needs to follow after fixing this

jamesmmchugh:
Yes, that's exactly what I think the behaviour should be also:

No templated variables, no secret = no error and output as-is

Templated variables, no secret = error (4xx)

Currently I'm working around it by putting all resources with secrets in a
different application in argocd, but it's not ideal in terms of project
structure.

vmwiz:
I'm getting the same kind of error

```
Error making API request. 
URL: GET https://<vault server>/v1/namespace
Code: 403. Errors: 
  * 1 error occurred: 
      * permission denied
```

The templated value is an environment variable inside a `Deployment` ressource.

[...]

PR #62, opened 2021-02-16
fix: Handle YAMLs with non-placeholder strings properly

jkayani:
### Description

Previously, the plugin would attempt to always get secrets from Vault for each manifest, even if there were no secrets in Vault for it. Now, the logic is as follows:

- If there are placeholders in the manifest template, try to get secrets from Vault, and return any appropriate error during that and during replacement
- Otherwise, skip reading from Vault, and emit the manifest as-is during replacement

Fixing this also introduced a new problem with `Secret`'s, since things under `data` need to be in base64 format, whereas everything else doesn't. So we add the following logic:

- If a `Secret.data` key has a `<placeholder>` value, we emit it's replacement base64 encoded. The expectation therefore is that the user puts data into Vault exactly as they want it loaded _into their app_
- Otherwise, we emit it as-is, meaning the user should hard-code any `Secret.data` values as base64 encoded. This makes for a proper GitOps experience

Anything in Secret that isn't in `data` is also emitted as-is (`metadata`, `secretData` fields)

**Fixes:** #54 

### Checklist

### Type of Change

- [x] Bugfix
- [ ] Feature
- [x] New tests
- [ ] Other (please describe)

### DIFF
--- a/pkg/kube/util.go
+++ b/pkg/kube/util.go
@@ -12,6 +12,55 @@ import (
 	k8yaml "k8s.io/apimachinery/pkg/util/yaml"
 )
 
+var placeholder, _ = regexp.Compile(`(?mU)<(.*)>`)
+
+// replaceableInner recurses through the given map and returns true if _any_ value is a `<placeholder>` string
+func replaceableInner(node *map[string]interface{}) bool {
+	obj := *node
+
+	for _, value := range obj {
+		valueType := reflect.ValueOf(value).Kind()
+		if valueType == reflect.Map {
+			inner, ok := value.(map[string]interface{})
+			if !ok {
+				panic(fmt.Sprintf("Deserialized YAML node is non map[string]interface{}"))
+			}
+			if replaceableInner(&inner) {
+				return true
+			}
+		} else if valueType == reflect.Slice {
+			for _, elm := range value.([]interface{}) {
+				switch elm.(type) {
+				case map[string]interface{}:
+					{
+						inner := elm.(map[string]interface{})
+						if replaceableInner(&inner) {
+							return true
+						}
+					}
+				case string:
+					{
+						if placeholder.Match([]byte(elm.(string))) {
+							return true
+						}
+					}
+				default:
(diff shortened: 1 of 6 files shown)


jkayani:
The PR description makes me think we should add some docs about this to the README. Thoughts?

pkg/kube/util.go
```diff
@@ -12,6 +12,55 @@ import (
...
+// replaceableInner recurses through the given map and returns true if _any_ value is a `<placeholder>` string
+func replaceableInner(node *map[string]interface{}) bool {
+	obj := *node
+
+	for _, value := range obj {
+		valueType := reflect.ValueOf(value).Kind()
+		if valueType == reflect.Map {
+			inner, ok := value.(map[string]interface{})
+			if !ok {
+				panic(fmt.Sprintf("Deserialized YAML node is non map[string]interface{}"))
```
werne2j, line 26: Are we using `panic` now instead of returning an error back to the generate method?

jkayani: No, that is what the current code does in `replaceInner` as well. I copied most of that to create this method. We can change it if you want, but since it's non-recoverable I just left it as-is

werne2j: As long as we are consistent with the error handling. I'm also not a fan of printing out the stack trace and i believe panic will do that. Maybe not in argocd as the error does not get logged but as a cli tool

jkayani: > I'm also not a fan of printing out the stack trace and i believe panic will do that

That's right, it does print the stack trace. But honestly in these kinds of scenarios, if the plugin fails because of this problem I think the stack trace would help debug the problem. For now I say we keep it, and maybe leave it as a pre 1.0 item to replace with returning `error`'s instead

jkayani: Will fix as part of: https://github.com/IBM/argocd-vault-plugin/issues/63

werne2j (approved):
Looks good!

Issue #66, opened 2021-02-22
Improve testing


werne2j:
We currently sit around 50% code coverage, we should look to try and bolster our test suite. Some areas that need to be improved but not the only areas to target:

- Github Auth
- Kubernetes Auth
- cmd/utils
- cmd/generate

jhoover4:
I can take this one too. Or at least update some of the tests.

werne2j:
@jhoover4 go ahead! We could always use improved testing

PR #90, opened 2021-03-09
tests: Add e2e tests for generate command and test vault

werne2j:
### Description
Add tests for the generate command

**In support of:** #66 

### Checklist

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [x] New tests
- [ ] Other (please describe)

### DIFF
--- a/pkg/kube/util_test.go
+++ b/pkg/kube/util_test.go
@@ -1,6 +1,7 @@
 package kube
 
 import (
+	"encoding/json"
 	"errors"
 	"reflect"
 	"testing"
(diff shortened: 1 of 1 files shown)


pkg/kube/util.go
```diff
@@ -155,6 +155,10 @@ func stringify(input interface{}) string {
 		{
 			return strconv.FormatBool(input.(bool))
 		}
+	case json.Number:
```
jkayani, line 158: Interesting, can you elaborate on how this is distinct from `Int`? I guess this code is written to assume data has been deserialized out of JSON into a struct so I don't understand what this is about

werne2j: This was necessary because of this https://github.com/IBM/argocd-vault-plugin/pull/90/files#diff-ef862b67a6c63e2e22fa26eb548c5c57f61388219beb2d7fee6ca1733af6542dR12.. i had a number in Vault and when it tried to do the replacement the type of the interface was `json.Number` instead of `Int` so i added this case to fix that error. The error was: 
```
panic: interface conversion: interface {} is json.Number, not string [recovered]
	panic: interface conversion: interface {} is json.Number, not string
```

jkayani: Hmmn OK

jkayani (approved):
Thanks this will be really helpful for catching regressions across releases

Issue #103, opened 2021-03-24
Argo CD - directory: recurse: true plugin bug 


JoshDanielWalker:
**Describe the bug**
When adding the plugin to Argo cd Application YAML if directory, recurse is set to true like below an error is given. 
`directory:
      recurse: true`
      
`multiple application sources defined: Directory,Plugin`

werne2j:
This error is because you are trying to define multiple plugins. You would need to remove `directory: recurse: true`. Or only use `Directory` this is not an issue with the plugin but how ArgoCD works as it only allows one "application source" at a time

JoshDanielWalker:
Thank you for your quick reply.
ok, I thought this might be an issue with Argo cd plugins in general.

JoshDanielWalker:
Removing directory: worked however, the templating syntax for the plugin picked up on some prometheus alert yaml.

 "< +Inf ) >".

eg `((( sum(container_memory_usage_bytes{image!="",container!="POD", namespace!="kube-system"}) by (container,pod,namespace)  / sum(container_spec_memory_limit_bytes{image!="",container!="POD",namespace!="kube-system"}) by (container,pod,namespace) ) * 100 ) < +Inf ) > 75`

JoshDanielWalker:
If there is whitespace could the plugin ignore the < > ignore?

jkayani:
@JoshDanielWalker Alternatively, what if we just an annotation to the YAML to tell AVP to skip it, something like `avp_ignore: True`?

JoshDanielWalker:
@jkayani definitely, this would help and would stop Argo from getting stuck in our case. 
I don't think whitespace should be included in the templating personally but I understand it is more complex to add this change?

jkayani:
> I don't think whitespace should be included in the templating personally but I understand it is more complex to add this change?

You're probably right, it should not. I think we can try adding a fix for that too - should just be a matter of changing the regex. But the annotation would be the proper way to tell AVP to ignore a file since that will work regardless of what the strings in your YAML look like

jkayani:
@JoshDanielWalker It's not particularly complex at all - it's just changing the regexp we use to recognize `<placeholder>`'s. 

However, I talked it over with @werne2j and we decided to omit adding that change in behavior because it'd be a big breaking change and the `avp_ignore` annotation is sufficient to address this use-case.

jkayani:
[...]

PR #104, opened 2021-03-24
feat: Add avp_ignore annotation to ignore YAMLs

jkayani:
### Description

- Adds an annotation to tell AVP to skip certain YAML files that may use the placeholder syntax for other purposes; i.e, less than or greater than sign

**Fixes:** #103 

### Checklist

### Type of Change

- [ ] Bugfix
- [x] Feature
- [x] New tests
- [ ] Other (please describe)


pkg/kube/template.go
```diff
@@ -43,13 +44,23 @@ func NewTemplate(template map[string]interface{}, backend types.Backend, prefix
 		kvVersion = kv
 	}
 
-	// Only worry about getting Vault secrets for templates with <placeholder>'s
-	replaceable := replaceableInner(&template)
-	var data map[string]interface{}
-	if replaceable {
-		data, err = backend.GetSecrets(path, kvVersion)
+	var avpIgnore bool
```
werne2j, line 47: I think we can set this to false on initialization on not worry about the err in converting the string to bool

pkg/kube/template.go
```diff
@@ -43,13 +44,23 @@ func NewTemplate(template map[string]interface{}, backend types.Backend, prefix
...
 		if err != nil {
-			return nil, err
+			avpIgnore = false
+		}
+	}
+
+	var data map[string]interface{}
+	var replaceable bool
+	if !avpIgnore {
+		replaceable = replaceableInner(&template)
```
werne2j, line 58: Instead of doing
```
var replaceable bool
if !avpIgnore {
  replaceable = replaceableInner(&template)
  if replaceable {
```
You could do
```
if replaceableInner(&template)
```

pkg/kube/template.go
jkayani: Actually bool default is false so really I can just ignore the `err`

pkg/kube/template.go
jkayani: I need to store the result of `replaceableInner`  for line 71 where we return the constructed `Template`

jkayani: And would prefer to not run `replaceableInner` twice since right now it runs through the whole file looking for `<thing>`'s

werne2j: Sure, should have looked further down. It just looks a little funky with the 3 different handlings of the bool, not a blocker

pkg/kube/template_test.go
```diff
@@ -77,536 +75,34 @@ func TestNewTemplate(t *testing.T) {
 			t.Fatalf("template does contain <placeholders> so GetSecrets should be called")
 		}
 	})
-}
-
-func TestToYAML_Deployment(t *testing.T) {
```
werne2j, line 82: Did you mean to delete all these tests?

jkayani: Ah shoot, was wondering why the coverage dropped so much

werne2j (approved):
Nice!

PR #96, opened 2021-03-15
feat!: Remove base64 encoding

werne2j:
### Description
Remove the base64 encoding that is currently happening on the fly and change expectations that use provide base64 in Vault if they want to use the data field of a secret.

**This is a breaking change**

**Fixes:** #92 

### Checklist
Please make sure that your PR fulfills the following requirements:
- [x] Reviewed the guidelines for contributing to this repository
- [x] The commit message follows the [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [x] Tests for the changes have been updated
- [x] Docs have been added / updated

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] New tests
- [ ] Build/CI related changes
- [ ] Documentation content changes
- [ ] Other (please describe)

Issue #58, opened 2021-02-09
Vault kv2 secret versioning support


ls-devon-hakel-kinko:
Does this plugin support [vault secret versioning](https://learn.hashicorp.com/tutorials/vault/versioned-kv)?
If not, do you plan on supporting it in the future?

Example:
```
kind: Secret
apiVersion: v1
metadata:
  name: example-secret
  annotations:
    vault_secret_version: '5'  # ability to specify secret revision
    avp_path: "path/to/secret"
type: Opaque
data:
  password: <password-vault-key>

```

werne2j:
@ls-devon-hakel-kinko we do not support secret versioning at this time. But we can discuss it to see if we can include it in the roadmap

jkayani:
I actually haven't used Vault kvv2 in much detail, but is it possible to set the secrets in Vault to their values in a given revision; i.e, rollback a Vault path to version X? Because if you could do that, then you could use the plugin as you desire.

jkayani:
I looked into it: https://www.vaultproject.io/docs/commands/kv/rollback and it seems you can do the following:

- Create version 1 of a Vault secret
- Create version 2
- Rollback to 1, which creates a version 3 with version 1's data

At that point running the plugin would fetch the version 3 (version 1 data) of the secret, effectively doing what you wanted. Is that enough?

llavaud:
I think it could be a usefull feature, and I see a use case:
we coud have several app that use the same secret path but for some reason for a specific app we want it to use a different version of the secret.
maybe a syntax like this: `<path:path/to/secret#key#version>`

jkayani:
Vault uses numeric versions: https://learn.hashicorp.com/tutorials/vault/versioned-kv

IBM SM uses UUIDs: https://cloud.ibm.com/apidocs/secrets-manager#get-secret-version

AWS SM has some sort of versioning too: https://docs.aws.amazon.com/secretsmanager/latest/userguide/getting-started.html#term_version

GCP SM as well: https://cloud.google.com/secret-manager/docs/managing-secret-versions#secretmanager-get-secret-version-api

So I guess it's possible to have this feature for all supported backends

jkayani:
This feature is available in https://github.com/IBM/argocd-vault-plugin/releases/tag/v1.3.0. See docs for details: https://ibm.github.io/argocd-vault-plugin/howitworks/

PR #113, opened 2021-04-19
Various Features

Aenima4six2:
### Description
**Support base64 encoded templates (ie: Output of kustomize secret generator)**

Given a Vault secret located at `kv/data/somesecret` with the following content:
```
{ "foo": "foo-value", "bar":"bar-value" }
```

and the following `kustomization.yaml` file in a directory named `example`...
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

secretGenerator:
  - name: test
    literals:
      - 'foo=<foo>'
      - 'bar=<bar>'
    options:
      annotations:
        avp_path: 'kv/data/foo'
```

Running `kustomize build example`, would produce the following secret. Note, it is not possible to output non base64 encoded content to `stringData` with kustomize.
```
apiVersion: v1
data:
  bar: PGJhcj4=
  foo: PGZvbz4=
kind: Secret
metadata:
  annotations:
    avp_path: kv/data/foo
  name: test-k29d8kmc86
type: Opaque
```

When rendered with `argocd-vault-plugin generate`, the following secret would be produced:
```
apiVersion: v1
data:
  bar: YmFyLXZhbHVl # bar-value
  foo: Zm9vLXZhbHVl # foo-value
kind: Secret
metadata:
  annotations:
    avp_path: kv/data/foo
  name: test-k29d8kmc86
type: Opaque
```

**Add Vault path query string support (ie: kv/data/foo?version=2)**
```
apiVersion: v1
kind: Secret
metadata:
  name: test
  annotations:
    avp_path: "kv/data/somesecret?version=7" # Explicitly fetch version 7 of the secret
type: Opaque
stringData:
    foo: <Some-Key-From-Version-7>
```

**Add support to auto template all available secret keys (via annotation avp_all_keys_to: data|stringData)**

Given a Vault secret located at `kv/data/somesecret` with the following content:
```
{ "foo": "foo-value", "bar":"bar-value" }
```

And the following `Secret`
```
apiVersion: v1
kind: Secret
metadata:
  annotations:
    avp_path: kv/data/somesecret
    avp_all_keys_to: stringData  # Include all keys in kv/data/somesecret
  name: test
type: Opaque
```

The following secret would be produced
```
apiVersion: v1
kind: Secret
metadata:
  annotations:
    avp_path: kv/data/somesecret
    avp_all_keys_to: stringData  # supports stringData or data
  name: test
stringData:
  foo: "foo-value"
  bar: "bar-value"
type: Opaque
```

### Checklist

### Type of Change
- [ ] Bugfix
- [x] Feature
- [x] New tests
- [ ] Other (please describe)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ bin
 .chglog
 test
 cover.out
+.idea
--- a/pkg/backends/vault.go
+++ b/pkg/backends/vault.go
@@ -3,6 +3,7 @@ package backends
 import (
 	"errors"
 	"fmt"
+	"net/url"
 
 	"github.com/IBM/argocd-vault-plugin/pkg/types"
 	"github.com/hashicorp/vault/api"
@@ -37,7 +38,14 @@ func (v *Vault) Login() error {
 
 // GetSecrets gets secrets from vault and returns the formatted data
 func (v *Vault) GetSecrets(path, kvVersion string) (map[string]interface{}, error) {
-	secret, err := v.VaultClient.Logical().Read(path)
+	parsed, err := url.Parse(path)
+	if err != nil {
+		return nil, err
+	}
+
+	data := parsed.Query()
+	parsed.RawQuery = ""
+	secret, err := v.VaultClient.Logical().ReadWithData(parsed.String(), data)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/kube/template.go
+++ b/pkg/kube/template.go
@@ -32,7 +32,8 @@ func NewTemplate(template map[string]interface{}, backend types.Backend, prefix 
 		return nil, fmt.Errorf("ToYAML: could not convert replaced template into %s: %s", obj.GetKind(), err)
 	}
 
-	path := fmt.Sprintf("%s/%s", prefix, strings.ToLower(obj.GetKind()))
+	kind := strings.ToLower(obj.GetKind())
+	path := fmt.Sprintf("%s/%s", prefix, kind)
 
 	annotations := obj.GetAnnotations()
 	if avpPath, ok := annotations["avp_path"]; ok {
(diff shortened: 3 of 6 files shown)


werne2j:
Hi @Aenima4six2 Thanks for the PR, could you please split this PR up to have one per new feature. That way each feature will be easier to review and discuss independently.

Aenima4six2:
> Hi @Aenima4six2 Thanks for the PR, could you please split this PR up to have one per new feature. That way each feature will be easier to review and discuss independently.

Hi @werne2j , I should have mentioned, I was planning on breaking this up and adding docs for each feature, if these additions seems like they are something that will be accepted into the repo. I'm happy to break things up if the intent is to move forward.

werne2j:
@Aenima4six2 From my perspective, i believe the ability to specify the version makes sense as we have an open issue #58 related.

I am trying to understand your first change `Support base64 encoded templates` a little more, if you could maybe help me to understand the need. 

`Add support to auto template all available secret keys` this seems like a feature request that we could potentially look at in the future. Right now we are focused with trying to get 1.0.0 released. So i would say you can create an issue and a PR and we can discuss it more there. 

Would like to hear what @jkayani thinks on these

Aenima4six2:
> @Aenima4six2 From my perspective, i believe the ability to specify the version makes sense as we have an open issue #58 related.
> 
> I am trying to understand your first change `Support base64 encoded templates` a little more, if you could maybe help me to understand the need.
> 
> `Add support to auto template all available secret keys` this seems like a feature request that we could potentially look at in the future. Right now we are focused with trying to get 1.0.0 released. So i would say you can create an issue and a PR and we can discuss it more there.
> 
> Would like to hear what @jkayani thinks on these

So for the `Support base64 encoded templates`, this just adds support for Secrets who's template is "pre" base64 encoded. Essentially, I'd like to use a Kustomize Secret Generators with the argocd-vault-plugin. Since kustomize secret generators always output the secret data (the template in this case) as base64 encoded strings in the `data` field, I can't currently use  argocd-vault-plugin templates in these secrets. 

`Add support to auto template all available secret keys` was a requirement for my use case. We currently have an ETL utility that projects, transforms and templates keys from various disparate vault secrets into aggregate, read only Vault secrets. We do this to centralize and offload responsibility templating and retrieving disparate secret data from each consumer. Given that, our consumers need not know about each key in the secret, only that they need everything in the secret.

jkayani:
Hi,

First @Aenima4six2 thanks for the PR! Here is what I understand from reading over it:

- You have a use-case to generate Secrets dynamically based on whatever is in Vault at deploy time. This plugin only works on YAMLs with `<placeholder>`'s already defined. Therefore, you have added an annotation to tell the plugin to generate `key: <key>`'s in the `Template`  based off what was in Vault, and then do its normal replacement work. This adds an assumption that the keys of the `Secret.data` match the keys in Vault exactly (which isn't required for normal AVP use right now)

- You have added explicit Vault versioning support. No disagreement with @werne2j this is a good add and at the least, I'd think that part can be accepted now as-is

- Kustomize will only allow for base64 output when generating Secrets. Therefore if you want to use Kustomize and then run AVP afterward, you need AVP to be able to decode base64'd `<placeholder>` and do it's normal work

Cool stuff. The major problems/questions I have are:

- https://github.com/IBM/argocd-vault-plugin/blob/8ae007d67c4a24082a09c3d15545652859d8d319/pkg/kube/util.go#L132 - in a previous PR https://github.com/IBM/argocd-vault-plugin/pull/96/files we'd decided that data in Vault must be encoded however it needs to be in the target k8s resource. So to replace keys in `Secret.data`, the values in Vault should already be base64'd and not in plain-text. This appears to conflict with that

- The dynamic Secret generation is clever but seems to beat around the bush: couldn't we just add the data to the k8s resource when given that annotation? I agree with @werne2j this would be a major feature so regardless we'll need to hold on that. Let's think about it 

- RE: the Kustomize workaround (3rd bullet above) - this seems like it could work, but I'd like to verify there is no way to reverse the operations (run AVP and then use Kustomize to build the YAMLs) before committing to a feature like this

Correct me if my understanding is wrong on the first 3 things. Happy to discuss further

werne2j:
@Aenima4six2 Lets start with the vault secret versioning. Create a separate PR for that change. And then if you could create issues for the other proposals and we can continue the conversation there. Will also help us gauge if there is interest from any other users on this issues.

PR #115, opened 2021-04-28
chore: merge next into main to prepare for 1.0.0 release

werne2j:
### Description
Merge `next` into `main` to prepare to create the 1.0.0 GA release

### Checklist
Please make sure that your PR fulfills the following requirements:
- [x] Reviewed the guidelines for contributing to this repository
- [x] The commit message follows the [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [x] Tests for the changes have been updated
- [x] Docs have been added / updated

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] New tests
- [ ] Build/CI related changes
- [ ] Documentation content changes
- [ ] Other (please describe)

PR #122, opened 2021-05-10
fix: Set VAULT_ variables from Viper and display replace errors better

jkayani:
### Description

Fixes the other problems discovered in testing #115:

- Sets `VAULT_` variables from Viper-read config file or k8s secret
- Prints the replacement errors newline separated to make them easier to read

### Checklist

### Type of Change

- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [x] New tests
- [ ] Other (please describe)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -155,7 +155,7 @@ func readConfigOrSecret(secretName, configPath string, v *viper.Viper) error {
 		}
 	}
 
-	for k, v := range viper.AllSettings() {
+	for k, v := range v.AllSettings() {
 		if strings.HasPrefix(k, "vault") {
 			var value string
 			switch v.(type) {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -155,14 +155,14 @@ func readConfigOrSecret(secretName, configPath string, v *viper.Viper) error {
 		}
 	}
 
-	for k, v := range v.AllSettings() {
+	for k, viperValue := range v.AllSettings() {
 		if strings.HasPrefix(k, "vault") {
 			var value string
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -155,7 +155,7 @@ func readConfigOrSecret(secretName, configPath string, v *viper.Viper) error {
 		}
 	}
 
-	for k, v := range viper.AllSettings() {
+	for k, v := range v.AllSettings() {
```
werne2j, line 158: Can we either change the `v` in the loop or `v` for viper just so it reads a little better?

jkayani: Oh duh right, no variable shadowing. Too bad we never added a linter :)

pkg/config/config.go
```diff
@@ -155,14 +155,14 @@ func readConfigOrSecret(secretName, configPath string, v *viper.Viper) error {
 		}
 	}
 
-	for k, v := range v.AllSettings() {
```
werne2j: I think you replaced the wrong `v`

jkayani: https://github.com/IBM/argocd-vault-plugin/pull/122/commits/143b13dc9fff6580239b4701140452a8d2dfb681#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R158

`viperValue` is the value for the given Viper key, seems right to me?

werne2j: I guess i got confused. Since `v` is the viper instance. And then `viperValue` is the viper settings. I guess it's not a big deal.

Issue #14, opened 2020-11-13
Add logic to create github releases 


werne2j:
We want to add to the github actions to be able to create a github release on some event

PR #135, opened 2021-06-08
ci: add release automation

teejaded:
### Description
Add release automation using goreleaser.  

Example release in my fork: https://github.com/teejaded/argocd-vault-plugin/releases/tag/v1.1.0

To trigger a release, create a new tag with the desired version number.

**Fixes:** https://github.com/IBM/argocd-vault-plugin/issues/14

### Checklist

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### Other information

I also updated the `argocd-vault-plugin version` command. Here's an example of the new output:

```bash
./dist/argocd-vault-plugin_darwin_amd64/argocd-vault-plugin version
argocd-vault-plugin v1.1.0 (5c126d26abe6dbb5f13bbeec6fe90ad39dd18554) BuildDate: 2021-06-08T18:52:31Z
```

For the release automation to work, please add a [secret GH_PAT](https://github.com/IBM/argocd-vault-plugin/settings/secrets/actions) with the value of a GitHub Personal Access Token that has repo scope.

See goreleaser doc for more information if needed.  
https://goreleaser.com/quick-start/

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,4 +28,4 @@ jobs:
           version: latest
           args: release --rm-dist
         env:
-          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,31 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      -
+        name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+      -
+        name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          # either 'goreleaser' (default) or 'goreleaser-pro'
+          distribution: goreleaser
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -24,4 +24,4 @@ release:
 changelog:
   filters:
     exclude:
-      - "^(ci|style|test)(\\([^)]+\\))?(!)?:"
+      - "^(ci|style|test|tests)(\\([^)]+\\))?(!)?:"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,27 @@
+project_name: "argocd-vault-plugin"
+
+builds:
+  - binary: "argocd-vault-plugin"
+    flags: "-trimpath"
+    ldflags: >-
+      -s -w
+      -X "github.com/IBM/argocd-vault-plugin/version.Version={{.Version}}"
+      -X "github.com/IBM/argocd-vault-plugin/version.BuildDate={{.Date}}"
+      -X "github.com/IBM/argocd-vault-plugin/version.CommitSHA={{.Commit}}"
+    env:
+      - "CGO_ENABLED=0"
+    goarch:
+      - amd64
+      - arm
+      - arm64
+    tags:
+      - netgo
+      - static_build
+
+release:
+  prerelease: auto
+
+changelog:
+  filters:
+    exclude:
+      - "^(ci|style|test)(\\([^)]+\\))?(!)?:"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -5,7 +5,7 @@ builds:
     flags: "-trimpath"
     ldflags: >-
       -s -w
-      -X "github.com/IBM/argocd-vault-plugin/version.Version={{.Version}}"
+      -X "github.com/IBM/argocd-vault-plugin/version.Version={{.Tag}}"
       -X "github.com/IBM/argocd-vault-plugin/version.BuildDate={{.Date}}"
       -X "github.com/IBM/argocd-vault-plugin/version.CommitSHA={{.Commit}}"
     env:
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ cover.out
 # Project workspaces
 .idea
 .vscode
+dist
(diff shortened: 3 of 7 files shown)


.goreleaser.yml
```diff
@@ -0,0 +1,27 @@
...
+      - netgo
+      - static_build
+
+release:
+  prerelease: auto
+
+changelog:
+  filters:
+    exclude:
+      - "^(ci|style|test)(\\([^)]+\\))?(!)?:"
```
jkayani, line 27: Will want to add `tests` too

.goreleaser.yml
```diff
@@ -0,0 +1,27 @@
+project_name: "argocd-vault-plugin"
+
+builds:
+  - binary: "argocd-vault-plugin"
+    flags: "-trimpath"
+    ldflags: >-
+      -s -w
+      -X "github.com/IBM/argocd-vault-plugin/version.Version={{.Version}}"
```
jkayani, line 8: It looks like this will do some code generation to set the variables in `version.go` accordingly? If so, cool. Though it's not obvious how the git tag is interpolated as `.Version` - maybe I need to read the docs for this `goreleaser`

teejaded: It's a feature of the golang linker.  https://golang.org/cmd/link/

```
-X importpath.name=value
	Set the value of the string variable in importpath named name to value.
	This is only effective if the variable is declared in the source code either uninitialized
	or initialized to a constant string expression. -X will not work if the initializer makes
	a function call or refers to other variables.
	Note that before Go 1.5 this option took two separate arguments.
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,31 @@
...
+      -
+        name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          # either 'goreleaser' (default) or 'goreleaser-pro'
+          distribution: goreleaser
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
```
werne2j, line 31: From my understanding you can use `GITHUB_TOKEN` instead of creating a new PAT https://docs.github.com/en/actions/reference/authentication-in-a-workflow#using-the-github_token-in-a-workflow

teejaded:
> One thing is i would like to keep the format of release docs that we have now. Today we use `git-chglog` to generate the release notes, would be nice if we can still do that and then specify the notes generated from that as the actual release notes

Do you want to continue to manually commit the CHANGELOG.md?

I can extract the release notes out of that file.

.goreleaser.yml
jkayani: Learned something new, thanks. And I honestly didn't now the delimitter for the Golang templates could be changed so I just used RegExp 🤷 . But like you said it worked out nice since we also have the specific-path-in-placeholder syntax too.

Sorry if I'm being dense but why does it say `version.Version={{.Version}}` and not `version.Version={{.Tag}}`?

teejaded: I'll switch it to `.Tag`.  I didn't know about that one when I wrote this because it wasn't mentioned in their release documentation.  I only found the link above after you asked about it!

teejaded:
Alternatively, you can commit your `.chglog` config and I can use git-chglog to generate the release notes.

https://goreleaser.com/customization/release/#custom-release-notes

Something like:
`goreleaser --release-notes <(git-chglog "{{.Tag}}")`

Something funny though is when I look at the git-chglog releases they're using goreleaser and the default release notes format from it.

werne2j:
@teejaded once we get 1.1.0 released, i will push up the `.chglog` config file. I would also like the release notes commited to `CHANGELOG.md`. Unless @jkayani would still like to see the changelog pushed up before the release is made?

jkayani:
@werne2j I think it's fine. Whenever we make a release I just want to read over the README and changelog which I can do by just reading the state of `master` before you tag and push

werne2j:
@teejaded to make things easier, here is the `git-chlog` config and template i use.

config.yaml
```yaml
style: github
template: CHANGELOG.tpl.md
info:
  title: CHANGELOG
  repository_url: https://github.com/IBM/argocd-vault-plugin
options:
  commits:
    filters:
      Type:
        - feat
        - fix
        - perf
        - refactor
  header:
    pattern: "^(\\w*)\\:\\s(.*)$"
    pattern_maps:
      - Type
      - Subject
  notes:
    keywords:
      - BREAKING CHANGE
```

CHANGELOG.tpl.md
```md
{{ range .Versions }}
<a name="{{ .Tag.Name }}"></a>
## {{ if .Tag.Previous }}[{{ .Tag.Name }}]({{ $.Info.RepositoryURL }}/compare/{{ .Tag.Previous.Name }}...{{ .Tag.Name }}){{ else }}{{ .Tag.Name }}{{ end }} ({{ datetime "2006-01-02" .Tag.Date }})

{{ range .CommitGroups -}}
### {{ .Title }}

{{ range .Commits -}}
* {{ .Subject }}
{{ end }}
{{ end -}}

{{- if .NoteGroups -}}
{{ range .NoteGroups -}}
### {{ .Title }}

{{ range .Notes }}
{{ .Body }}
{{ end }}
{{ end -}}
{{ end -}}
{{ end -}}
```

.goreleaser.yml
```diff
@@ -0,0 +1,27 @@
...
+    goarch:
+      - amd64
+      - arm
+      - arm64
+    tags:
+      - netgo
+      - static_build
+
+release:
+  prerelease: auto
```
werne2j, line 22: This should be set back to false, we dont normally do pre releases

werne2j:
If possible, i would also like to publish to homebrew, looks like goreleaser makes it pretty easy

.goreleaser.yml
teejaded: I saw some old release candidates so I thought it might be useful.  I think all this does is if you check the prerelease checkbox, it will still run the automation.  I don't think it does anything otherwise.

teejaded:
> If possible, i would also like to publish to homebrew, looks like goreleaser makes it pretty easy

https://goreleaser.com/customization/homebrew/

Okay I looked into it.  It needs a `homebrew-tap` repo.  
This seems available, but I imagine it might need to be shared across projects: https://github.com/IBM/homebrew-tap

Here's two example tap repos:
https://github.com/goreleaser/homebrew-tap
https://github.com/fluxcd/homebrew-tap

werne2j:
Thanks for looking into that @teejaded, i guess we can hold off on homebrew and just focus on releases. I will create another issue to do more investigating on homebrew

werne2j:
Closing in favor of https://github.com/IBM/argocd-vault-plugin/pull/157

Issue #124, opened 2021-05-14
Support decryption values.yaml


lcc3108:
My English is not so fluent
**Is your feature request related to a problem? Please describe.**
argocd-vault-plugin don't support decryption values.yaml

**Describe the solution you'd like**
I think support decryption values.yaml  more easier to use

**Describe alternatives you've considered**

**Additional context**

## situation
values.yaml
```yaml
password: <path:secret/data/harbor#password> 
```
templates/secret.yaml
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: {{ .Values.name }}
type: Opaque
data:
  PASSWORD: {{ .Values.password | b64enc | quote }}
```

key : "secret/data/harbor#password"  
value: "password-at-vault"

key and value set at vault 
## if support values.yaml decryption
values.yaml 
decrypt by argocd-vault-plugin
```yaml
password: "password-at-vault"
```
templates/secret.yaml
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: {{ .Values.name }}
type: Opaque
data:
  PASSWORD: "cGFzc3dvcmQtYXQtdmF1bHQ="
  #cGFzc3dvcmQtYXQtdmF1bHQ= equal base64 -d "password-at-vault"
```
but now release
values.yaml
```yaml
password: <path:secret/data/harbor#password>
```
templates/secret.yaml
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: {{ .Values.name }}
type: Opaque
data:
  PASSWORD: PHBhdGg6c2VjcmV0L2RhdGEvaGFyYm9yI3Bhc3N3b3JkPiA=
  #PHBhdGg6c2VjcmV0L2RhdGEvaGFyYm9yI3Bhc3N3b3JkPiA= equal base64 -d "<path:secret/data/harbor#password> "
```
---
It is difficult to change many files under the above circumstances.

werne2j:
> apiVersion: v1
kind: Secret
metadata:
  name: {{ .Values.name }}
type: Opaque
data:
  PASSWORD: {{ .Values.password | b64enc | quote }}

Do you have control of this secret? If so, you can just remove the ` | b64enc | quote` and either use the `base64encode` modifier we provide of just input a bas64d value in vault.

lcc3108:
It is also possible to change the Helm template.

[...]

PR #138, opened 2021-06-09
fix: implicitly base64 encode base64 secret values

teejaded:
### Description
If a secret is successfully decoded from base64, assume it needs to be re-encoded.  This allows for secret values with multiple placeholders or concatenation with other strings. 

**Fixes:** https://github.com/IBM/argocd-vault-plugin/issues/124

### Checklist

### Type of Change

- [x] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,20 @@ fieldRef:
   some-credential: <path:somewhere/in/my/vault#credential>
 ```
 
+In addition to the default behavior, the plugin will attempt to decode base64 encoded strings in Secrets to look for placeholders.  If a placeholder is found, it is replaced and the resulting string is re-base64 encoded.  In most cases it is not necessary to use the `base64encode` modifier in this scenario. In the following example the plugin will decode the `POSTGRES_URL` value, find the template `postgres://<username>:<password>@<host>:<port>/<database>?sslmode=require`, and base64 encode it after replacement.
+
+```yaml
+kind: Secret
+apiVersion: v1
+metadata:
+  name: example-secret
+  annotations:
+    avp.kubernetes.io/path: "path/to/secret"
+type: Opaque
+data:
+  POSTGRES_URL: cG9zdGdyZXM6Ly88dXNlcm5hbWU+OjxwYXNzd29yZD5APGhvc3Q+Ojxwb3J0Pi88ZGF0YWJhc2U+P3NzbG1vZGU9cmVxdWlyZQ==
+```
+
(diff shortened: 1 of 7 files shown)


werne2j:
> If a secret is successfully decoded from base64, assume it needs to be re-encoded.

Are we sure about this assumption?

jkayani:
Recap of how the plugin works now:

- `stringData`, placeholder alone: replace as-is, works

- `stringData`, placeholder part of a larger plain string: replace as-is, works since entire result is plain-text

- `data`, placeholder alone: fails as expected since replacement is plain-text. Workaround is to use `stringData` or base64 data in Vault.

- `data`, placeholder part of a larger string: fails as expected since entire result is plain-text. Workaround is to use `stringData` or load in Vault the larger string pre-replaced, base64-encoded and then only keep the `placeholder` in the `data`. 

- `data`, placeholder with base64 modifier: base64 the replacement, works

- `data`, placeholder with base64 modifier pre-encoded as base64: decode placeholder, base64 replacement, works

- `data`, placeholder with base64 modifier part of a larger plain string pre-encoded as base64: fails, since we only base64 the replacement and not the whole string. The current workaround would be to put the entire plain string as desired to appear in `data` in Vault and just keep a placeholder with the base64 modifier. Or to put the entire string as desired to appear in base64 format in Vault, and just keep a placeholder alone in `data`.

This PR addresses the last one so that `data` placeholders (possibly part of a larger string) pre-encoded as base64 will be replaced and the entire result is re-encoded as base64. You wouldn't need the `base64` modifier here at all I think.

I think it's a good feature to have but it will be inconsistent with how the plugin works in the 3rd and 4th scenarios

jkayani:
So if we're going down this route then to make 3rd and 4th scenarios consistent, we'd have to change the rules so that AVP will re-encode into base64 the result of any replacements of strings in `data` (making the base64 modifier unnecessary). That would a breaking change

jkayani (approved):
Talked to Jake some more about it and Jake pointed out that the 3rd and 4th scenarios I brought up would only happen if a YAML outside of someone's control (so from an upstream chart) did something like:

```yaml
data:
  credentials: |
     [block]
      password: {{ .values.password }}
```

and just expected that to work, no base64'ing done anywhere. That's unreasonable. 

If the user did have control over the YAML, this is easily solved by swapping `data` for `stringData`. So there is nothing "inconsistent" about what this PR does and I approve!

Sorry about the delay :)

werne2j:
@teejaded Can you add documentation to the readme about how the encoding/decoding will work?

PR #141, opened 2021-06-10
fix: Set proper aws default region

werne2j:
### Description
Still had a reference to `AVP_AWS_REGION`, added aws region to readme under aws section

### Checklist

### Type of Change

- [x] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ metadata:
 type: Opaque
 fieldRef:
 
-  # So, <key> is NOT a placeholder
+  # So, <KEY> is NOT a placeholder
   description: 'Selects a field of
     the pod: supports metadata.name,
     metadata.namespace, `metadata.labels[''<KEY>'']`,
@@ -117,7 +117,7 @@ metadata:
 type: Opaque
 fieldRef:
 
-  # So, <key> is NOT a placeholder
+  # So, <KEY> is NOT a placeholder
   description: 'Selects a field of
     the pod: supports metadata.name,
     metadata.namespace, `metadata.labels[''<KEY>'']`,
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ metadata:
 type: Opaque
 fieldRef:
 
-  # So, <key> is NOT a placeholder 
+  # So, <key> is NOT a placeholder
   description: 'Selects a field of
     the pod: supports metadata.name,
     metadata.namespace, `metadata.labels[''<KEY>'']`,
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -91,7 +91,7 @@ metadata:
 type: Opaque
 fieldRef:
 
-  # So, <key> is NOT a placeholder 
+  # So, <key> is NOT a placeholder
   description: 'Selects a field of
     the pod: supports metadata.name,
     metadata.namespace, `metadata.labels[''<KEY>'']`,
```
jkayani, line 97: Not relevant to this PR at all but can you change `<key>` to `<KEY>` so the example is more clear. I got the casing wrong

jkayani (approved):
Makes me wonder if we should just forward all `AWS_` env variables to the AWS SDK like we do for Vault

werne2j:
> Makes me wonder if we should just forward all AWS_ env variables to the AWS SDK like we do for Vault

I actually think it will do that already. If the aws go sdk read using `os.getenv` and you set an appropriate env var, it should get read. We could remove the code to set the region all together and just rely on env vars if you think that is a better approach

werne2j:
For example:
https://github.com/aws/aws-sdk-go/blob/main/aws/defaults/defaults.go
https://github.com/aws/aws-sdk-go/blob/main/aws/credentials/env_provider.go

jkayani:
Should clarify I was thinking about this: https://github.com/IBM/argocd-vault-plugin/blob/main/pkg/config/config.go#L167 where we collect configs for Vault from all the different places into Viper, and then set the appropriate Vault ones as env variables to send them to the SDK. That way we don't have to fiddle with Vault options in code. 

> We could remove the code to set the region all together and just rely on env vars if you think that is a better approach

I think that's what I'm thinking, yes. Have a loop like just for Vault, but looking for `AWS` prefixed Vault keys and just set them as env variables and let the AWS SDK handle using them

README.md
```diff
@@ -112,12 +112,12 @@ metadata:
...
 
   # Notice, `avp.kuberenetes.io/ignore` annotation is set
-  annotations: 
+  annotations:
     avp.kuberenetes.io/ignore: "true"
 type: Opaque
 fieldRef:
 
-  # So, <key> is NOT a placeholder 
+  # So, <key> is NOT a placeholder
```
jkayani, line 120: I am the worst - can you get this one too?

jkayani (approved):
LGTM otherwise

werne2j:
@jkayani i will create an issue to passthrough all aws env vars like vault and then merge this so we can get 1.1.0 release

Issue #148, opened 2021-06-16
Authentication for AWS secret manager via IAM roles


fvdnabee:
It is my understanding that currently support for authentication when using the AWS Secrets Manager as a back end is limited to just the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables (at least this is what is documented). 

For pods running on an AWS k8s cluster (EKS), it is possible to authenticate with AWS using IAM roles via [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). As documented [here](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html#pod-configuration), EKS sets the following environment variables for pods configured to employ IRSA: `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE`. One nice thing about IRSA is that you only need to add a single annotation to the serviceaccount that runs your pod (if you have setup the IAM roles), there is no need to fiddle with access keys.

Would it be possible to authenticate with AWS from the argocd-vault-plugin via IAM/IRSA? Could it be that IAM auth is already available, as the plugin relies on the aws go SDK for authentication? I've been unable to test this.

werne2j:
Hi @fvdnabee as of right now, `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are required. We don't have anything on the roadmap for adding any other AWS auth methods right now. 

However, we are accepting pull requests if you want suggest a change! It would be nice to nativly be able to accept all aws auth methods, so it might be as simple as removing the checks around the id and secret for the `awssecretsmanager` type.

fvdnabee:
@werne2j from my testing it appears that removing https://github.com/IBM/argocd-vault-plugin/blob/main/pkg/config/config.go#L112 is sufficient indeed. The SDK will try a chain of auth method: see https://github.com/aws/aws-sdk-go#configuring-credentials It does appear wise to ensure the proper region is set, currently the code expects an env variable to be present to set the region (default to `us-east-2` if it absent). However, I'm not sure whether this is the recommended method of setting the region; as e.g. the shared credentials file can also specify the region.

I'll do some further testing and try to submit a PR later today.

fvdnabee:
Closed with merging #151

PR #151, opened 2021-06-16
fix: env variables check blocks many AWS authentication methods 

fvdnabee:
### Description

Enforcing the specification of the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` env variables blocks other authentication methods supported by the AWS go SDK from being used by argocd-vault-plugin. Even within env vars authentication, there exist other options than just `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`: e.g. using an assumed role on ec2/EKS. This PR removes the unnecessary check for the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` env vars. It does remain recommended to set the proper AWS region via an environment variables however. 

**Fixes:** https://github.com/IBM/argocd-vault-plugin/issues/148

### Checklist

### Type of Change

- [x] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### Other information

Note, this PR was tested successfully with an IAM role for service accounts as the auth method on an AWS EKS cluster. In this test the following env variables were set by the EKS Pod Identity Webhook:
```
AWS_ROLE_ARN=arn:aws:iam::xxx:role/argocd-repo-server-secretsmanager-my-cluster
AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
AWS_DEFAULT_REGION=eu-west-1
AWS_REGION=eu-west-1
```

Note, when no credentials are found the user is shown the following error after this PR is applied:
```
AVP_TYPE=awssecretsmanager ./bin/argocd-vault-plugin_1.2.0dev_linux_amd64 generate /tmp/secret.yaml
Error: NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
Usage:
  argocd-vault-plugin generate <path> [flags]

Flags:
  -c, --config-path string   path to a file containing Vault configuration (YAML, JSON, envfile) to use
  -h, --help                 help for generate
  -s, --secret-name string   name of a Kubernetes Secret containing Vault configuration data in the argocd namespace of your ArgoCD host (Only available when used in ArgoCD)

NoCredentialProviders: no valid providers in chain. Deprecated.
	For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -109,13 +109,6 @@ func New(v *viper.Viper, co *Options) (*Config, error) {
 		}
 	case types.AWSSecretsManagerbackend:
 		{
-			if !v.IsSet(types.EnvAWSAccessKey) || !v.IsSet(types.EnvAWSSecretAccessKey) {
-				return nil, fmt.Errorf("Must provide %s and %s for backend type %s",
-					types.EnvAWSAccessKey,
-					types.EnvAWSSecretAccessKey,
-					types.AWSSecretsManagerbackend,
-				)
-			}
 			s := session.Must(session.NewSession(&aws.Config{
 				Region: aws.String(v.GetString(types.EnvAWSRegion)),
 			}))
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -153,7 +153,7 @@ func readConfigOrSecret(secretName, configPath string, v *viper.Viper) error {
 	}
 
 	for k, viperValue := range v.AllSettings() {
-		if strings.HasPrefix(k, "vault") {
+		if strings.HasPrefix(k, "vault") || strings.HasPrefix(k, "aws") {
 			var value string
 			switch viperValue.(type) {
 			case bool:
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -109,9 +109,13 @@ func New(v *viper.Viper, co *Options) (*Config, error) {
 		}
 	case types.AWSSecretsManagerbackend:
 		{
-			s := session.Must(session.NewSession(&aws.Config{
+			s, err := session.NewSession(&aws.Config{
 				Region: aws.String(v.GetString(types.EnvAWSRegion)),
-			}))
+			})
+			if err != nil {
+				return nil, err
+			}
+
 			client := secretsmanager.New(s)
 			backend = backends.NewAWSSecretsManagerBackend(client)
 		}
--- a/pkg/types/constants.go
+++ b/pkg/types/constants.go
@@ -2,20 +2,18 @@ package types
 
 const (
 	// Environment Variable Constants
-	EnvAvpType            = "AVP_TYPE"
-	EnvAvpRoleID          = "AVP_ROLE_ID"
-	EnvAvpSecretID        = "AVP_SECRET_ID"
-	EnvAvpAuthType        = "AVP_AUTH_TYPE"
-	EnvAvpGithubToken     = "AVP_GITHUB_TOKEN"
-	EnvAvpK8sRole         = "AVP_K8S_ROLE"
-	EnvAvpK8sMountPath    = "AVP_K8S_MOUNT_PATH"
-	EnvAvpK8sTokenPath    = "AVP_K8S_TOKEN_PATH"
-	EnvAvpIBMAPIKey       = "AVP_IBM_API_KEY"
-	EnvAvpKvVersion       = "AVP_KV_VERSION"
-	EnvAvpPathPrefix      = "AVP_PATH_PREFIX"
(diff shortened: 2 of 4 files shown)


werne2j:
@fvdnabee thanks for the PR! Can you please rebase your branch onto main?

fvdnabee:
> @fvdnabee thanks for the PR! Can you please rebase your branch onto main?

Done!

werne2j:
@fvdnabee Everything looks good, could you add a test or two to cover the new logic?

fvdnabee:
@werne2j checking the error value of `session.NewSession` is of limited use as the aws sdk doesn't actually validate credentials at that time. You'll only know whether your credentials are valid and whether they are authorized  when performing API calls (at least this is my understanding). 

That said, I've added a test for a corner case in bc1982f which is covered by the SDK. Another test case demonstrates web identity federation authentication.

werne2j (approved):
Looks good!

pkg/config/config.go
```diff
@@ -109,16 +109,13 @@ func New(v *viper.Viper, co *Options) (*Config, error) {
...
 		{
-			if !v.IsSet(types.EnvAWSAccessKey) || !v.IsSet(types.EnvAWSSecretAccessKey) {
-				return nil, fmt.Errorf("Must provide %s and %s for backend type %s",
-					types.EnvAWSAccessKey,
-					types.EnvAWSSecretAccessKey,
-					types.AWSSecretsManagerbackend,
-				)
-			}
-			s := session.Must(session.NewSession(&aws.Config{
+			s, err := session.NewSession(&aws.Config{
```
jkayani, line 112: This looks fine. The only problem here is that if someone is setting their AVP config through a k8s Secret, or a config file, then specifying the `AWS_BLAH` parameters will not get sent to the AWS SDK, and it won't. Perhaps you can add some code to loop through the Viper settings and set all `AWS_` config keys as env variables like we do for `VAULT_`: https://github.com/IBM/argocd-vault-plugin/blob/main/pkg/config/config.go#L159

pkg/types/constants.go
```diff
@@ -2,20 +2,18 @@ package types
...
+	EnvAvpSecretID     = "AVP_SECRET_ID"
+	EnvAvpAuthType     = "AVP_AUTH_TYPE"
+	EnvAvpGithubToken  = "AVP_GITHUB_TOKEN"
+	EnvAvpK8sRole      = "AVP_K8S_ROLE"
+	EnvAvpK8sMountPath = "AVP_K8S_MOUNT_PATH"
+	EnvAvpK8sTokenPath = "AVP_K8S_TOKEN_PATH"
+	EnvAvpIBMAPIKey    = "AVP_IBM_API_KEY"
+	EnvAvpKvVersion    = "AVP_KV_VERSION"
+	EnvAvpPathPrefix   = "AVP_PATH_PREFIX"
+	EnvAWSRegion       = "AWS_REGION"
```
jkayani, line 16: If you decide to take the other suggestion then you can drop this variable too

jkayani (commented):
LGTM, but this change will only apply to AVP config from env variables. I'd prefer to have it work regardless of source of configuration, thus my comments. Let me know your thoughts

pkg/config/config.go
fvdnabee: Unsure of the use-cases here as I wouldn't add AWS credentials to an AVP config file/secret on k8s, preferring IAM auth or envFrom from a dedicated AWS secret instead. But others might see a use for it, so I'll add support for setting envs vars for each `AWS_` entry in the config file.

pkg/types/constants.go
fvdnabee: For most auth methods it's important to ensure a valid region is set though. I believe an env variable is the most common way of doing so. In light of this I'd  suggest to keep EnvAWSRegion as a dedicated constant, see also PR #155 .

jkayani: I'm convinced

pkg/config/config.go
fvdnabee: @jkayani I just pushed e846f41 with a unit test for `AWS_` config entries. Hope I understood you right.

fvdnabee: Note, the fake credentials are taken from https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html#envvars-set

jkayani (approved):
Exactly what I wanted to see, thanks a lot @fvdnabee !

Issue #162, opened 2021-07-07
Placeholder replacement - replaces the whole value of the key. Not only placeholder.


rimasgo:
Hello!

**Is your feature request related to a problem? Please describe.**
I am trying to configure alert-manager (Prometheus) on Openshift V4 cluster and the configuration for it is not very GitOps friendly (https://prometheus.io/docs/alerting/latest/configuration/). At first glance it looks okay however the essentially it's a huge HELM values file and it has a password (smtp_user_password) hence it has to be created as a secret.

I have come up with a way of creating it with named templates (helm) however the place holders overrides the whole value of the key and also only the last place holder is picked up.

This is how my secret YAML looks before running generate with Vault plugin (edited)

apiVersion: v1
kind: Secret
metadata:
  name: alertmanager-main
  namespace: openshift-monitoring
  annotations:
    avp.kubernetes.io/path: "secret/data/namespace/alertmanager"
data:
[...]

PR #163, opened 2021-07-07
fix: dont replace entire string if base64 modifier included

werne2j:
### Description
Fixes an issue where if the bas64 modifier is included in a longer string it was replacing the entire string

**Fixes:** #162 

### Checklist

### Type of Change

- [x] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/pkg/kube/template.go
+++ b/pkg/kube/template.go
@@ -69,7 +69,7 @@ func (t *Template) Replace() error {
 	replaceInner(&t.Resource, &t.TemplateData, replacerFunc)
 	if len(t.replacementErrors) != 0 {
 		errMessages := make([]string, len(t.replacementErrors))
-		for idx, err := range(t.replacementErrors) {
+		for idx, err := range t.replacementErrors {
 			errMessages[idx] = err.Error()
 		}
 		return fmt.Errorf("Replace: could not replace all placeholders in Template:\n%s", strings.Join(errMessages, "\n"))
--- a/fixtures/input/nonempty/secret_embedded_base64.yaml
+++ b/fixtures/input/nonempty/secret_embedded_base64.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  annotations:
+    avp.kubernetes.io/kv-version: "1"
+    avp.kubernetes.io/path: secret/testing
+  name: embedded-secret-base64
+  namespace: default
+type: Opaque
+data:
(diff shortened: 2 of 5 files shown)


pkg/kube/template.go
```diff
@@ -66,10 +66,10 @@ func (t *Template) Replace() error {
 		replacerFunc = genericReplacement
 	}
 
-	replaceInner(&t.Resource, &t.TemplateData, replacerFunc)
+	replaceInner(&t.Resource, &t.Resource.TemplateData, replacerFunc)
```
werne2j, line 69: @jkayani can you take close look here? Not sure how it worked before since `t` doesn't have a `TemplateData` field, its in the Resource struct

jkayani: It's just Go struct embedding right? If struct X has a struct Y, you can access Y's fields via X: https://golang.org/doc/effective_go#embedding. So your change is fine I think but it should also be fine unchanged

pkg/kube/template.go
```diff
@@ -79,7 +79,7 @@ func (t *Template) Replace() error {
 
 // ToYAML seralizes the completed template into YAML to be consumed by Kubernetes
 func (t *Template) ToYAML() (string, error) {
-	res, err := yaml.Marshal(&t.TemplateData)
+	res, err := yaml.Marshal(&t.Resource.TemplateData)
```
jkayani, line 82: If I'm right that applies here too

pkg/kube/template.go
werne2j: Oh yeah, that makes sense. I will revert to limit the scope of changes

jkayani (approved):
I think I understand the fix now:

Original problem is that base64 modifier was initially implemented using the `nonStringReplacment` escape hatch which replaces the whole string with just the replacement, causing the issue reported in GH.  https://github.com/IBM/argocd-vault-plugin/blob/4426843da2be71489b30efc539fb84cf8cb1ca36/pkg/kube/util.go#L114 and https://github.com/IBM/argocd-vault-plugin/blob/4426843da2be71489b30efc539fb84cf8cb1ca36/pkg/kube/util.go#L136

This solution avoids that by returning a byte slice of base64 chars of the original string in the `ReplaceAll` which avoids the problem of removing the rest of the string of the placeholder: https://github.com/IBM/argocd-vault-plugin/blob/50e0e07664a187e997ecaa109ae55a83e965dfbc/pkg/kube/util.go#L114

Issue #165, opened 2021-07-14
Allow Empty Keys to Be Bypassed Without Error


jhoover4:
**Is your feature request related to a problem? Please describe.**
My use case is that we store all of our secrets in one vault config for each environment. For some of these environments (say dev), a key may not have been created and the existence should be optional. Using the plugin makes having a value for these keys required.

**Describe the solution you'd like**
I'd like to have an annotation that can be used to configure the default behavior and not error out if a key is missing, instead merely give a warning.

**Describe alternatives you've considered**
I haven't thought of other solutions but open to ideas!

Perhaps our way of keeping these keys in vault is not ideal, but I do think I'd rather us not have to have blank values in vault for some of these keys (or have to keep track of it).

**Additional context**
Example of current behavior of the plugin that I'd like to change:

![Screen Shot 2021-07-14 at 12 21 56 PM](https://user-images.githubusercontent.com/20507620/125665389-1528ccc2-51ae-4121-a01e-cc714494b258.png)

I'd also be happy to contribute a PR for this, but I wanted to make sure the feature is okay with the maintainers first or if there's a reason it may not be needed.

jkayani:
@jhoover4 Interesting. When you say:

> I'd like to have an annotation that can be used to configure the default behavior and not error out if a key is missing, instead merely give a warning.

What should happen with the placeholder? I assume it should just remain there as-is? I think the behavior you're asking for is certainly doable, something like `avp.kubernetes.io/ignore-missing-placeholders: true`. I'd be curious as to @werne2j 's thoughts on this. 

Also I'm not sure how useful that warning would be (at least in the context of ArgoCD) since to my knowledge the stderr of an ArgoCD plugin is never displayed in the UI (but maybe it appears in the repo-server pod logs? not sure). If this is a use-case not involving ArgoCD then my point is moot here.

jhoover4:
I was thinking to just skip the placeholder altogether. Either that or we can have a null value we always use?

Yeah I think not having a warning would make sense, it would probably just be junk if you knew what you were doing.

You do see the stderr by the way if the plugin errors out. Its available when you look at the errors under the application.

werne2j:
@jhoover4 

> I was thinking to just skip the placeholder altogether.

If by this you mean to leave the placeholder in the file we can do that, however, if you have:
```
apiVersion: v1
kind: Secret
metadata:
  name: example-secret
  annotations:
    avp.kubernetes.io/path: project/sets/data/test-secret
[...]

PR #177, opened 2021-08-10
feat: add allow-empty annotation.

jhoover4:
### Description
PR to address feature change requested in #165.

Open to suggestions/review!

### Checklist

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/pkg/kube/util_test.go
+++ b/pkg/kube/util_test.go
@@ -368,7 +368,7 @@ func TestGenericReplacement_allowEmpty(t *testing.T) {
 		},
 	}
 
-	assertFailedReplacement(&dummyResource, &expected, t)
+	assertSuccessfulReplacement(&dummyResource, &expected, t)
 }
 
 func TestSecretReplacement(t *testing.T) {
--- a/pkg/kube/util_test.go
+++ b/pkg/kube/util_test.go
@@ -3,7 +3,7 @@ package kube
 import (
 	"encoding/base64"
 	"encoding/json"
-	"errors"
+	"fmt"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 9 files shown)


werne2j:
@jhoover4 Looks like the tests are failing. Can you look into that? Thanks!

jhoover4:
@werne2j whoops sorry about that! Looks like something got messed up in a merge. Its fixed now.

pkg/kube/util_test.go
```diff
@@ -329,6 +329,48 @@ func TestGenericReplacement_missingValue(t *testing.T) {
...
+		},
+		Data: map[string]interface{}{
+			"namespace": "default",
+		},
+		Annotations: map[string]string{
+			"avp.kubernetes.io/allow-empty": "true",
+		},
+	}
+
+	assertFailedReplacement(&dummyResource, &expected, t)
```
jkayani, line 371: Why are we asserting failed here? If I read this right, the presence of the new annotation makes it so that `replacementErrors` is left empty if a placeholder can't be replaced. With that and the expected and actual data matching, this test seems wrong to me.

jkayani:
This looks fine to me otherwise but I want to make sure we have a valid test - otherwise we'll have problems later on with catching regressions

pkg/kube/util_test.go
jhoover4: Ah I just copied a test above, I think both assertions would work cause we're just trying to assert that there are no errors. I can push a change making it `assertSuccessfulReplacement` if thats clearer? Or should I make a new assertion?

werne2j:
@jhoover4 After looking it over more, not sure if the test is enough. Could you also add an example YAML as part of the end to end test. You can see the test i am referring to here https://github.com/IBM/argocd-vault-plugin/blob/main/cmd/generate_test.go#L66. 

Essentially, add a YAML to `fixtures/input/nonempty` and then add the expected outcome to `fixtures/output/all.yaml`. Will take a try or two to get the right order but i think it will be a good test.

jhoover4:
Okay fair enough! I'll work on an e2e test tomorrow morning and get a new commit in.

pkg/kube/util_test.go
jkayani: > I can push a change making it assertSuccessfulReplacement

Yes, this would be better and looks like you did that. Thanks!

jhoover4:
Okay guys, I feel like I was kinda phoning it in before 😬. Took a closer look at this and I think the changes I just pushed are what my team really needs. We need the keys to be remove in a Secret or ConfigMap if they are missing. 

I've updated the annotation name to better reflect this and have restricted the annotation to Secrets and ConfigMaps only, since I believe any others would simply break the manifest.

Also added two e2e tests as requested, one testing the feature and another testing the error if any other manifest outside Secret or ConfigMap is tried with the annotation.

Let me know if all this makes sense or if I need to do some more work!

werne2j:
@jhoover4 Thanks for the update, we will take a look and discuss the proposed feature

jkayani (approved):
The feature and tests look to be implemented well. I'll leave it up to @werne2j to make the final decision. I appreciate you updating the docs as well.

werne2j:
@jhoover4 the only thing i'd like to see is example yamls added to the generate test. For reference: https://github.com/IBM/argocd-vault-plugin/pull/177#issuecomment-897097321 This will do a full test and not just start at the `replace`. Thanks!

jhoover4:
> @jhoover4 the only thing i'd like to see is example yamls added to the generate test. For reference: [#177 (comment)](https://github.com/IBM/argocd-vault-plugin/pull/177#issuecomment-897097321) This will do a full test and not just start at the `replace`. Thanks!

Ah sorry I misunderstood. I think I got it now though! Let me know if what I added works.

werne2j (approved):
Looks good! Thanks @jhoover4

PR #182, opened 2021-08-19
feat: return plain secret value if secret ist not a valid json

mms-gianni:
### Description
It is possible to save plaintext secrets in aws secretsmanager. (Example use case : License). Since they are not a JSON object, the need to be treated differently. 

<path:/path/to/my/secret#plainValue>

In case the secret is not a JSON it will return the plain secret value. The key needs to be "#plainValue" 

### Checklist

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/pkg/backends/awssecretsmanager.go
+++ b/pkg/backends/awssecretsmanager.go
@@ -40,9 +40,15 @@ func (a *AWSSecretsManager) GetSecrets(path string, _ map[string]string) (map[st
 	var dat map[string]interface{}
 
 	if result.SecretString != nil {
-		err := json.Unmarshal([]byte(*result.SecretString), &dat)
-		if err != nil {
-			return nil, err
+		if json.Valid([]byte(*result.SecretString)) {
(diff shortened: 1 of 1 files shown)


werne2j:
@mms-gianni Thanks for this PR. At this time, i am not sure we want to introduce this feature as it will only work in certain scenarios. We are open to another approach to handle this and it is something we are looking into with other secret managers. The workaround for now is to put your License, Cert, etc. in a Key/Value pair in the secret. This will allow you to pull those values and inject them into your yaml files.

mms-gianni:
@werne2j Thank you for looking into my PR. 

As for AWS-Secretmanager there are only these two scenarios: 
- The AWS-Secret is a JSON object
- The AWS-Secret is plaintext

Since it is implemented in the Secretsmanager backend it will not affect other backends or other scenarios.

Putting them in a JSON Object is not an option for us. We have other Application expecting a plaintext value and have already too many plaintext secrets configured.

Issue #153, opened 2021-06-17
Lack of permission scopes for production setup, everyone creating a secret can create any secret in vault


ksiadz:
**Describe the bug**
There are no permissions scopes in Argo using one single k8s SA or one single AppRole. Meaning that even if you have proper Vault policies with scopes the ArgoCD plugin is still using one Vault Role which forces a practice of one role, assuming most people use Argo to deploy apps of multiple teams, one bad actor in any team can ask to create a secret in his k8s namespace which would be a security issue.

I would expect something like vault-agent which assumes different k8s SA to access different secrets and sends SA_name and NS_name to the vault which does policy check based on those.

Something like
```yaml
kind: Secret
apiVersion: v1
metadata:
  name: example-secret
  namespace: <k8s-namespace in which allowed SA is based>
  annotations:
    avp.kubernetes.io/path: "path/to/secret"
    avp.kubernetes.io/sa-in-use: "sa-name"
type: Opaque
data:
  password: <password-vault-key>

```

I would also expect either Argo Application /ApplicationSet blocking deploying a secret in a namespace that is not a destination namespace in Argo App yaml or some other mechanism blocking unwanted secrets escalation.

werne2j:
Hi @ksiadz, thanks for the issue. After looking into this a workaround that can be done to help improve security. The first thing is that you can take advantage of the Argo CD Application YAML to specify environment variables. This way you can define different AppRole or K8s SA. So you would be able to do:

```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: some-resource
  namespace: some-ns
spec:
  destination:
    namespace: dest-ns
    name: server-name
  source:
    path: some/path
    repoURL: https://github.com/some/repo
    targetRevision: HEAD
    plugin:
      name: argocd-vault-plugin
      env:
        - name: AVP_AUTH_TYPE
           value: approle
        - name: AVP_ROLE_ID
           value: some-approle-id
        - name: AVP_SECRET_ID
           value: some-secret-id
  project: default
```

This would allow you to specify different role/secrets for each application. That way you dont have to have one app role that can access all secrets. 

> I would also expect either Argo Application /ApplicationSet blocking deploying a secret in a namespace that is not a destination namespace in Argo App yaml or some other mechanism blocking unwanted secrets escalation or actually build-in a way to use vault checks.

[...]

PR #184, opened 2021-08-23
feat: allow selecting the secret namespace

blairdrummond:
### Description

Remove the hard-coded `Secrets("argocd")` for fetching the configuration secret, and instead default to the namespace of the service-account. This is with the aim of supporting this plugin with the [ArgoCD operator](https://argocd-operator.readthedocs.io/en/latest/) so that you can have this plugin on multiple argocd instances in one cluster (with different Vault roles)

**Relates to:** https://github.com/IBM/argocd-vault-plugin/issues/153

**Note: this basically assumes that either there is a mounted service account or the user explicitly names `-s argocd/vault-secret` in order to specify the secret namespace. That could be seen as a regression, but a fallback to "argocd" could easily be added.**


jkayani (commented):
This change seems reasonable and I'm fine with the syntax change to specify a namespace and secret within. It _is_ a breaking change though, so we'll need to update the docs and I guess bump the major version. @werne2j any disagreement?

I also see we don't have any tests for the `-s` and `-c` options so I won't ask you to add any unless you'd like to. But we should probably do that at some point

jkayani (commented):
>  but a fallback to "argocd" could easily be added

Maybe that's preferable to avoid the breaking change

blairdrummond:
> > but a fallback to "argocd" could easily be added
> 
> Maybe that's preferable to avoid the breaking change

There is one *tiny* edgecase, where if they were previously using a service-account in another namespace (other than "argocd") but sourcing the secret from the argocd namespace, then the new default would be the service-account namespace instead of the previous behavior. So I think it would still be a breaking change, but it might break *less*

jkayani (commented):
Current behavior: `argocd` namespace is used all the time with `--secret-name`

Desired behavior: Namespace of service account token is used unless a specific namespace is given - in a backwards compatible way

Maybe this should take a different form:

`--service-account-namespace` to tell AVP to read the secret from the namespace of the service account

`--secret-name` to tell AVP the name of the secret to read. Can be in the form `<string>/<string>` to specify a different namespace than `argocd`

Valid uses:

- `generate --secret-name vault/credentials`: reads `credentials` secret from `-n vault`
- `generate --service-account-namespace --secret-name credentials`: reads `credentials` from the namespace of the service account
- `generate --secret-name credentials`: reads `credentials` from `-n argocd`

Invalid uses:
- `generate --service-account-namespace`: no secret name given
- `generate --service-account-namespace --secret-name vault/credentials`: 2 different namespace options given

This way the current behavior is preserved in the 3rd valid use-case and you get the flexibility you need. It's a bit more work but I think it's preferable. Thoughts?

@werne2j In case you have a better idea

blairdrummond:
@jkayani if you give a :+1: I can implement your proposal above. Let me know

werne2j:
Going back to the original request..
> This is with the aim of supporting this plugin with the ArgoCD operator so that you can have this plugin on multiple argocd instances in one cluster (with different Vault roles)

This can be done today.. As we support plain environment variables. So your instance of Argo CD can use `envFrom` and whatever secret name you want. Then you dont have to pass the `-s` command to the generate command.

The alternative is we also support a config file. So you could mount a secret to a path and then pass the `-c` flag to the generate command. 

Maybe i am misunderstanding the request but you should be able to have multiple instance of the plugin in one cluster.

blairdrummond:
Does the ArgoCD Operator currently support that? https://github.com/argoproj-labs/argocd-operator/issues/395 I don't know if that's in [the spec](https://github.com/argoproj-labs/argocd-operator/issues/395) yet

werne2j:
@blairdrummond Ah right.. the operator. That has caused more than one problem with lack of supported features. My main focus is not to introduce any breaking changes. It seems that solution @jkayani suggested would provide the functionality you are looking for with no breaking changes. So i am on board with that feature.

blairdrummond:
Ok sounds good, will implement the changes shortly

blairdrummond:
@jkayani I've implemented your proposal for the `--service-account-namespace`. Let me know what you think!

werne2j:
@blairdrummond Could you please add tests to cover the new lines aded?

jkayani (commented):
Looks good, just want to see some tests for these flags.

It looks we handle the case of passing `--service-account-namespace` without `--secret-name` by just letting the plugin fail due to missing configuration, which is fine with me. All the other cases look to be handled properly.

werne2j:
@blairdrummond are you still working on this pull request?

blairdrummond:
@werne2j yes! Sorry, it hasn't been priority for us lately but will be soon (a few weeks). If you need it sooner I can find someone on our team to look at it

werne2j:
No worries @blairdrummond was just checking before closing the PR. Appreciate your contibutions!

werne2j:
Closing this in favor of https://github.com/argoproj-labs/argocd-vault-plugin/pull/251

PR #183, opened 2021-08-22
feat: Add support for GCP Secret Manager

loxley:
### Description
Add GCP Secret Manager as a new backend

### Checklist
Please make sure that your PR fulfills the following requirements:
- [X] Reviewed the guidelines for contributing to this repository
- [X] The commit message follows the [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [X] Tests for the changes have been updated
- [X] Docs have been added / updated

### Type of Change

- [ ] Bugfix
- [X] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] New tests
- [ ] Build/CI related changes
- [ ] Documentation content changes
- [ ] Other (please describe)

Issue #159, opened 2021-06-29
Add support for other secret management tools


werne2j:
Currently we support HashiCorp Vault, IBM Secrets Manager and AWS Secrets Manager as backends. We should look into any other secret management tools that we can add support for. GCP Secrets Manager and Azure Key Vault are two candidates

werne2j:
Adds GCP Secret Manager https://github.com/IBM/argocd-vault-plugin/pull/183

jkayani:
GCP SM available in https://github.com/IBM/argocd-vault-plugin/releases/tag/v1.3.0

werne2j:
We can close this now. As we also have Azure in an open PR

PR #189, opened 2021-08-26
feat: Support placeholder replacements with versioned secrets

jkayani:
### Description

Major highlights:

- Refactors IBM SM backend to use their SDK instead of the Vault API. Originally this was because I wanted to support versioned secrets which requires using the SDK. Then I found that `arbitrary` secrets aren't versioned in IBM SM. So, this isn't strictly necessary but hopefully worth keeping
- Adds versioned secret support to AWS SM. No syntax changes needed
- Adds versioned secret support to GCP SM. This requires a syntax change from what was introduced in #183. See https://github.com/IBM/argocd-vault-plugin/issues/58#issuecomment-906480254
- Adds versioned secret support to Hashicorp Vault. ~There are no automated tests for this feature~
- Adds new annotation to specify secret version

Other things:

- Turn off logging on the in-memory test Vaults since they are not useful
- Fix the in memory test Vault to actually use KV-V2 backend
- Let Cobra print errors to `stderr` instead of doing it ourselves too, which leads to error messages printing twice

I will call out other things I did to make sure we're good with them as PR review comments. 

As of this PR, these are the supported syntax(es) for placeholders:
```
<generic>
<path:/specific/path/to/secret#key>
<path:/specific/path/to/secret#key#version>
```

**This PR is not done yet** - once we agree on the way it works and stuff, I will delete the old definition of `GetSecrets` from the `Backend` interface and rename `GetSecretsVersioned` to it. I will also update docs then only. 

**Fixes:** #58 

### Checklist

### Type of Change

- [ ] Bugfix
- [x] Feature
- [x] Refactoring (no functional changes, no api changes)
- [x] New tests
- [ ] Other (please describe)


loxley:
Hey guys, I think the addition of having versions is awesome, but I recently started to implement the Azure backend as well (take a look here: https://github.com/loxley/argocd-vault-plugin/tree/add-azure-keyvault ) and I found out that in Azure they  have the ability to version each secret like:

https://testkv.vault.azure.net/secrets/foo/8f8da2e06c8240808ee439ff093803b5

Where the hash at the end is the "version" of secret "foo". I have not implemented that yet in my branch above since I just realized they have this versioning. Maybe something to take into account? I plan on creating a PR for you but saw that you were refactoring it :)

fixtures/input/nonempty/secret_path.yaml
```diff
@@ -8,3 +8,4 @@ metadata:
 type: Opaque
 data:
   SECRET_VAR: <path:secret/testing#secret-var-value>
```
werne2j, line 10: Can you add a file to nonempty that is using the secret version annotations as well?

werne2j (commented):
Code looks good, one small thing to add for the tests. And you said you were doing a separate PR for tests?

jkayani:
> Code looks good, one small thing to add for the tests. And you said you were doing a separate PR for tests?

For _docs_ I will do a separate PR, yes. This PR already has tests

werne2j:
@jkayani I was referring to
> Can you add a file to nonempty that is using the secret version annotations as well?

jkayani:
@loxley 

> I have not implemented that yet in my branch above since I just realized they have this versioning. 

I appreciate you working on Azure KeyVault! I suggest you rebase your work onto `main` once this gets merged. Pertaining to your work, this PR will add a parameter to `GetSecrets` for the version and it's up to your implementation of the Azure backend to build the API call correctly given the version. We can discuss further in #159 if you want.

jkayani:
@werne2j Understood - there are tests verifying the annotation behavior but they don't take input all the way from a fixture file. That's what I meant by there are already tests - will add a fixture for this too.

werne2j (approved):
Looks great!

jkayani:
@werne2j I think I forgot to cover the case of specifying the secret version as an annotation and then using inline-path placeholders without a version at the end. In that case, the proper behavior should be to use the annotation-provided secret version but I think the code will just assume no version was specified since they are inline-placeholders. Let me add code/tests for this.

jkayani:
Discussed with @werne2j and actually it makes more sense for the current behavior - a secret annotation only applies to generic placeholders, and that inline-path placeholders specify a version if desired. Given that a big use-case for inline-path is to use _different_ secrets that are versioned independently, using the secret annotation for inline-paths without a version specifier makes little sense.

werne2j (approved):
Double approve

PR #193, opened 2021-08-31
feat: Add support for Azure Key Vault

loxley:
### Description
Add Azure Key Vault as new backend

### Checklist

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/docs/backends.md
+++ b/docs/backends.md
@@ -308,3 +308,61 @@ data:
   current-password-again: <path:projects/12345678987/secrets/test-secret#password#latest>
   password-old: <path:projects/12345678987/secrets/test-secret#password#another-version-id>
 ```
+
+### AZURE Key Vault
+
+##### Azure Authentication
+Refer to the [Use environment-based authentication](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization#use-environment-based-authentication) in the Azure SDK for Go.
+
+Any secrets that are disabled in the key vault will be skipped if found.
+
+For Azure, `path` is the unique name of your key vault.
+
+**Note**: Versioning is only supported for inline paths.
+
+These are the parameters for Azure:
+```
+AVP_TYPE: azurekeyvault
+```
+
+##### Examples
+
+###### Path Annotation
+
+```yaml
+kind: Secret
+apiVersion: v1
+metadata:
+  name: test-secret
+  annotations:
+    avp.kubernetes.io/path: "keyvault"
+type: Opaque
+data:
+  password: <test-secret>
+```
+
+###### Inline Path
+
+```yaml
+kind: Secret
+apiVersion: v1
+metadata:
+  name: test-secret
+type: Opaque
+data:
+  password: <path:keyvault#test-secret>
+```
+
+###### Versioned secrets
+
+```yaml
+kind: Secret
+apiVersion: v1
+metadata:
+  name: test-secret
+type: Opaque
+data:
+  current-password: <path:keyvault#password>
+  current-password-again: <path:keyvault#password#8f8da2e06c8240808ee439ff093803b5>
+  password-old: <path:keyvault#password#33740fc26214497f8904d93f20f7db6d>
+```
--- a/pkg/backends/azurekeyvault.go
+++ b/pkg/backends/azurekeyvault.go
@@ -1,0 +1,78 @@
+package backends
+
+import (
+	"context"
+	"fmt"
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/keyvault/keyvault"
+	"path"
+	"strings"
+	"time"
+)
+
+// AzureKeyVault is a struct for working with an Azure Key Vault backend
+type AzureKeyVault struct {
+	Client keyvault.BaseClient
+}
+
+// NewAzureKeyVaultBackend initializes a new Azure Key Vault backend
+func NewAzureKeyVaultBackend(client keyvault.BaseClient) *AzureKeyVault {
+	return &AzureKeyVault{
+		Client: client,
+	}
+}
+
+// Login does nothing as a "login" is handled on the instantiation of the Azure SDK
+func (a *AzureKeyVault) Login() error {
+	return nil
+}
+
+// GetSecrets gets secrets from Azure Key Vault and returns the formatted data
+// For Azure Key Vault, `kvpath` is the unique name of your vault
+func (a *AzureKeyVault) GetSecrets(kvpath string, version string, _ map[string]string) (map[string]interface{}, error) {
+	kvpath = fmt.Sprintf("https://%s.vault.azure.net", kvpath)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	data := make(map[string]interface{})
+
+	secretList, err := a.Client.GetSecrets(ctx, kvpath, nil)
+	if err != nil {
+		return nil, err
+	}
+	// Gather all secrets in Key Vault
+	for ; secretList.NotDone(); secretList.NextWithContext(ctx) {
+		for _, secretItem := range secretList.Values() {
+			secret := path.Base(*secretItem.ID)
+			if version == "" {
+				secretResp, err := a.Client.GetSecret(ctx, kvpath, secret, "")
+				if err != nil {
+					return nil, err
+				}
+				data[secret] = *secretResp.Value
+				continue
+			}
+			// In Azure Key Vault the versions of a secret is first shown after running GetSecretVersions. So we need
+			// to loop through the versions for each secret in order to find the secret that has the specific version.
+			secretVersions, _ := a.Client.GetSecretVersions(ctx, kvpath, secret, nil)
+			for ; secretVersions.NotDone(); secretVersions.NextWithContext(ctx) {
+				for _, secretVersion := range secretVersions.Values() {
+					// Azure Key Vault has ability to enable/disable a secret, so lets honour that
+					if !*secretVersion.Attributes.Enabled {
+						continue
+					}
+					// Secret version matched given version
+					if strings.Contains(*secretVersion.ID, version) {
+						secretResp, err := a.Client.GetSecret(ctx, kvpath, secret, version)
+						if err != nil {
+							return nil, err
+						}
+						data[secret] = *secretResp.Value
+					}
+				}
+			}
+		}
+	}
+
+	return data, nil
+}
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	cloud.google.com/go v0.81.0
+	github.com/Azure/azure-sdk-for-go v57.0.0+incompatible
 	github.com/IBM/go-sdk-core/v5 v5.5.0
 	github.com/IBM/secrets-manager-go-sdk v1.0.24
 	github.com/aws/aws-sdk-go v1.40.9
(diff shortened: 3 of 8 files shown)


loxley:
The error from the `ubuntu-latest` pipeline puzzles me a bit, it works as intended when run the tests locally with Go 1.16.7 .

pkg/backends/azurekeyvault.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// Login does nothing as a "login" is handled on the instantiation of the Azure SDK
+func (a *AzureKeyVault) Login() error {
+	return nil
+}
+
+// GetSecrets gets secrets from Azure Key Vault and returns the formatted data
+func (a *AzureKeyVault) GetSecrets(kvpath string, version string, _ map[string]string) (map[string]interface{}, error) {
+	kvpath = "https://" + kvpath + ".vault.azure.net"
```
jkayani, line 30: Not a big deal, but should `fmt.Sprintf` be used here instead?

jkayani:
Failing test:

```
=== RUN   TestNewConfigMissingParameter
    config_test.go:292: *backends.AzureKeyVault should not instantiate
--- FAIL: TestNewConfigMissingParameter (0.01s)
```

Will try to take a look myself in a bit

pkg/backends/azurekeyvault.go
loxley: Agree 👍 fixing

jkayani:
@loxley I'm not really sure what's going on either. I tried running this in the Docker image `golang:1.16.7 ` but couldn't reproduce it there either. I guess you should try adding some debugging logs to the test to figure out how the Azure SDK is succeeding with init if you're missing one of the parameters. 

@werne2j Thoughts on changing the Actions configuration to use Go 1.17 instead? Anyway I think that's what you and I are running locally so we might as well run the CI on that too? And maybe this error will solve itself?

loxley:
> @loxley I'm not really sure what's going on either. I tried running this in the Docker image `golang:1.16.7 ` but couldn't reproduce it there either. I guess you should try adding some debugging logs to the test to figure out how the Azure SDK is succeeding with init if you're missing one of the parameters.

I think I might know why! So, Github runners is actually running in Azure datacenters... as per:

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

Then it tries with the last auth option which is same as f.ex GCP and others use if you are running "inside" their datacenters.

loxley:
If I can't come up with a good solution to prevent MSI from init, perhaps I can remove that test? We would still have the `TestNewConfig` test where it checks if env variables are set correctly.

jkayani:
@loxley Good debugging - but then why does the test pass on the macos runners? Those also presumably run in Azure and should have the same env variables? 

Anyway I agree - if you can't work around the SDK, I think it's OK to omit that test. I don't think it's worth the effort to wrap stuff in an interface and try to mock it out just for this. At most I'd ask you to leave a comment in `config_test` explaining why this test was omitted.

jkayani:
> GitHub hosts Linux and Windows runners on Standard_DS2_v2 virtual machines in Microsoft Azure with the GitHub Actions runner application installed. 

> GitHub hosts macOS runners in GitHub's own macOS Cloud.

Oh, ignore my macos comment.

jkayani:
@loxley We can probably work around this by just containerizing the tests, right? Let me try making a PR for that and it works then we can go with the original testing strategy. Would be good to be able to reproduce test results closely as possible too.

loxley:
> @loxley We can probably work around this by just containerizing the tests, right? Let me try making a PR for that and it works then we can go with the original testing strategy. Would be good to be able to reproduce test results closely as possible too.

Agree, will you ping me when done?

pkg/backends/azurekeyvault.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Login does nothing as a "login" is handled on the instantiation of the Azure SDK
+func (a *AzureKeyVault) Login() error {
+	return nil
+}
+
+// GetSecrets gets secrets from Azure Key Vault and returns the formatted data
+func (a *AzureKeyVault) GetSecrets(kvpath string, version string, _ map[string]string) (map[string]interface{}, error) {
+	kvpath = fmt.Sprintf("https://%s.vault.azure.net", kvpath)
```
werne2j, line 31: I would like to maybe understand this logic a little more. I see 6 different `for-loops` and wonder if there is a way that some of that can be combined. I imagine the performance on this will not be ideal, especially if there are a lot of different inline paths.

werne2j: In an inline scenario, i imagine we can just get the secret by name and then output just that so we dont have to run through this logic for every inline secret

loxley: @werne2j Agree, I somewhere had an idea that it could be nice to use some metadata that is in Azure Key Vault and use it in some way perhaps. But you are right, if not a performance hit, at least a longer fetching time and the first 3 for loops can easily be removed. :+1:  Let me get back with an update to this PR after the weekend if its OK.

loxley: I have hopefully cleaned out the code enough now and also added documentation. Let me know what you think :)

jkayani:
@loxley FYI the CI has been modified to use Docker for Linux runs #197  - so now would be a good time to rebase and see if it works for this test case (the `linux` and `mac` quality checks)

pkg/backends/azurekeyvault.go
jkayani: So for Azure KV, the path isn't actually a `/` delimited path because otherwise you wouldn't have a valid URL there. Would you mind adding a comment to note this, something like:

```
// GetSecrets does blah blah blah
// For Azure KV, the `kvpath` is actually blah, because blah blah
```

I'm guessing the SDK will handle provide a helpful error message for errors caused by giving something like `/path/to/secret` for `kvpath`

This might be good to mention in the docs too - your call on that though, I'm not too familiar with Azure KV so maybe what I'm saying is obvious to a user of that service

pkg/backends/azurekeyvault.go
```diff
@@ -0,0 +1,77 @@
...
+				data[secret] = *secretResp.Value
+				continue
+			}
+			// In Azure Key Vault the versions of a secret is first shown after running GetSecretVersions. So we need
+			// to loop through the versions for each secret in order to find the secret that has the specific version.
+			secretVersions, _ := a.Client.GetSecretVersions(ctx, kvpath, secret, nil)
+			for ; secretVersions.NotDone(); secretVersions.NextWithContext(ctx) {
+				for _, secretVersion := range secretVersions.Values() {
+					// Azure Key Vault has ability to enable/disable a secret, so lets honour that
+					if !*secretVersion.Attributes.Enabled {
```
jkayani, line 60: Lets add this to the docs that we honor this attribute

jkayani (approved):
This LGTM aside from my small comments: secret versions aren't directly addressable through the SDK so if a specific version is desired we iterate through a list of them, skipping disabled ones, until we find the requested one. Otherwise it's a simple lookup. The code is a lot more clear now, thanks for the cleanup.

jkayani (commented):
But I want to be sure the tests pass after rebased on latest `main` first

werne2j:
My suggestion is to maybe leave off the secret versions right now (Unless you need it). As the version is scoped to one secret that has one value. Having an annotation for that secret version isn't very helpful. We can come back to it in the future. But I think this PR may be better off with just handling the base functionality first. @jkayani thoughts on this?

jkayani:
@werne2j It's true that Azure KV seems to version individual keys rather than secrets (our terms). But in that case specifying the version via inline-path is useful and it works. Maybe instead we can just document that using the annotation for Azure KV just doesn't make sense and will cause errors, like how we document using the annotation with Vault KV-V1 is useless.

pkg/backends/azurekeyvault.go
loxley: Updated text and in docs.

loxley:
I have rebased against main now and updated docs. Do you think its enough or do you want some text about annotations as you discussed?

docs/backends.md
```diff
@@ -308,3 +308,63 @@ data:
...
+###### Versioned secrets
+
+```yaml
+kind: Secret
+apiVersion: v1
+metadata:
+  name: test-secret
+  annotations:
+    avp.kubernetes.io/path: "keyvault"
+    avp.kubernetes.io/secret-version: "8f8da2e06c8240808ee439ff093803b5"
```
werne2j, line 363: If we are going to use versions, since we established that the annotation version doesn't make sense, I would say lets remove that from the example and then in the code lets ignore that annotation (It actually looks like we are already ignoring) and then add a note that says versioning only available for inline paths.

werne2j (commented):
Really appreciate the work @loxley there are just a couple thoughts i have on the code/docs.

pkg/backends/azurekeyvault.go
```diff
@@ -0,0 +1,78 @@
...
+	defer cancel()
+
+	data := make(map[string]interface{})
+
+	secretList, err := a.Client.GetSecrets(ctx, kvpath, nil)
+	if err != nil {
+		return nil, err
+	}
+	// Gather all secrets in Key Vault
+	for ; secretList.NotDone(); secretList.NextWithContext(ctx) {
```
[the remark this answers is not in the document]
josandstrom: Lets say you have a kubernetes object like:

```yaml
kind: Secret
apiVersion: v1
metadata:
  name: example-secret
type: Opaque
data:
  password: <path:josannkv#foo>
  password-2: <path:josannkv#josansecret>
  password-3: <path:josannkv#foo#8f8da2e06c8240808ee439ff093803b5>
  password-4: <path:josannkv#josansecret#33740fc26214497f8904d93f20f7db6d>
  password-5: <path:josannkv#foo#4509bc1bbc2341ca8e195b23f829af01>
```

1. So lets say the secret list consists of `josansecret` then followed by `foo`. 

2. Version `8f8da2e06c8240808ee439ff093803b5` gets passed in from password-3 above.

3. The for loop would fail since `josansecret` does not contain `8f8da2e06c8240808ee439ff093803b5`, but the next interation ( `foo` ) would have it but the for loop never gets to that point right breaking the secret lookup.

Or am I thinking wrong about this?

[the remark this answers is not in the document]
werne2j: Yeah, i guess you are right, I was thinking about it with the mindset of each inline doing a `GetSecret` with the version and not thinking about each secret in the list of secrets trying to retrieved with that version. 

Ideally, we would not need to loop the secrets but instead do a simple lookup to get one secret by name and version for the inlines paths. But i think the the project is written it doesn't allow for this right now.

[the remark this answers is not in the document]
josandstrom: Yeah, its always possible to improve that in the future and if so then also fix this secret <-> version mapping.

docs/backends.md
josandstrom: Yeah, its really confusing doing the above with how this is working. Let me add that to the documentation.

jkayani (approved):
Aside from Jakes comments which should be addressed, this LGTM. Thanks for all the work, especially given the weird "GH Actions is in Azure and that matters" scenario

docs/backends.md
josandstrom: Docs updated, hopefully its OK now.

werne2j (approved):
Thanks @loxley

PR #195, opened 2021-09-01
feat: add vault token option

blairdrummond:
### Description

Add option to authenticate with a Vault token directly. Mostly for
local testing. Let me know what your requirements are in terms of tests

### Checklist
- [X] Docs have been added / updated

### Type of Change

- [ ] Bugfix
- [X] Feature
- [ ] New tests
- [ ] Other (please describe)

### Other information

The primary use-case is for local or ad-hoc testing, where you do not necessarily have the more serious authentication methods set up. I.e. grab a temporary token from a UI, test out the plugin on your manifests locally, then push to the repo after.


jkayani (commented):
This seems good to me. 

At most it'd be nice to have a test that creates in an in-memory test Vault setup to accept some string token and have your test use this AuthType + the token to read from Vault. It seems like that is possible when running a dev server from the CLI: https://www.vaultproject.io/docs/commands/server#dev-options so I wonder if it's possible programatically via `vault.CoreConfig.DevToken`: https://github.com/hashicorp/vault/blob/v1.8.2/vault/core.go#L581. I might be misunderstanding what that struct field is. 

If that doesn't work or its way more complicated then I'm fine with this as-is.

werne2j:
My preference on this would be to use the native `VAULT_TOKEN` env var instead of specifying a new env var

blairdrummond:
> My preference on this would be to use the native `VAULT_TOKEN` env var instead of specifying a new env var

Done!

docs/config.md
```diff
@@ -43,9 +43,11 @@ We support all Vault Environment Variables listed [here](https://www.vaultprojec
 | Name            | Description | Notes |
 | --------------- | ----------- | ----- |
 | AVP_TYPE           | The type of Vault backend  | Supported values: `vault`, `ibmsecretsmanager`, `awssecretsmanager` and `gcpsecretmanager` |
+| VAULT_ADDR         | The Vault URL              | The Vault address (e.g. `http://0.0.0.0:8200`)                                             |
```
werne2j, line 46: We shouldn't need to add this here as above we link directly to the Vault Env Var list

pkg/auth/vault/token.go
```diff
@@ -0,0 +1,32 @@
...
+	tokenAuth := &TokenAuth{
+		AccessToken: token,
+	}
+
+	return tokenAuth
+}
+
+// Authenticate authenticates with Vault and returns a token
+func (t *TokenAuth) Authenticate(vaultClient *api.Client) error {
+
```
werne2j, line 24: I think this can be updated to actually just be no-op and return nil. Because the client will use the `VAULT_TOKEN` with no intervention. So i would remove the `NewTokenAuth` method, just return nil out of `Authenticate` and you can just make `TokenAuth` an empty struct.

pkg/config/config.go
```diff
@@ -86,6 +86,12 @@ func New(v *viper.Viper, co *Options) (*Config, error) {
 				} else {
 					return nil, fmt.Errorf("%s cannot be empty when using Kubernetes Auth", types.EnvAvpK8sRole)
 				}
+			case types.TokenAuth:
+				if v.IsSet(types.EnvAvpVaultToken) {
```
werne2j, line 90: You should be able to use the vault const here since its exported `api.EnvVaultToken` https://github.com/hashicorp/vault/blob/main/api/client.go#L42 (I think)

pkg/types/constants.go
```diff
@@ -7,6 +7,7 @@ const (
 	EnvAvpSecretID       = "AVP_SECRET_ID"
 	EnvAvpAuthType       = "AVP_AUTH_TYPE"
 	EnvAvpGithubToken    = "AVP_GITHUB_TOKEN"
+	EnvAvpVaultToken     = "VAULT_TOKEN"
```
werne2j, line 10: Should be able to remove this

werne2j (changes requested):

werne2j:
@blairdrummond could you please add tests to the config test file? Should be able to follow the pattern used for the other backends

pkg/config/config_test.go
```diff
@@ -226,6 +234,14 @@ func TestNewConfigMissingParameter(t *testing.T) {
 			},
 			"*backends.Vault",
 		},
+		{
+			map[string]interface{}{
+				"AVP_TYPE":      "vault",
+				"AVP_AUTH_TYPE": "token",
+				"VAULT_TOKEN":   "token",
```
werne2j, line 241: This test is to make sure the backend does not instantiate, so you will want to leave off the VAULT_TOKEN

werne2j (commented):
Everything looks good, just one small change

pkg/config/config_test.go
blairdrummond: Done!

werne2j:
@blairdrummond Just need to update your branch to be up to date with main and we should be good to go

werne2j (approved):
Thanks @blairdrummond! Will give @jkayani a chance to take one more look.

jkayani (approved):
LGTM

PR #197, opened 2021-09-03
ci: Use Docker for CI on Linux

jkayani:
### Description

Dockerize the CI environment on Linux to avoid this problem: https://github.com/IBM/argocd-vault-plugin/pull/193#issuecomment-911876575. 

This solution works based on this Actions run https://github.com/IBM/argocd-vault-plugin/runs/3508366010?check_suite_focus=true which makes these changes atop PR 193 to prove we avoid the special-case of Azure auth solely via cloud-provider metadata endpoints

Specific CI changes:

- Split the CI matrix since we use Docker on Linux not on Mac
- Add a Dockerfile that downloads the Golang deps at build time
- Run `go vet` only on our code (nor our deps) and run `go test` in a network isolated container to avoid successful Azure auth
- Upload code coverage only once
- Remove some un-used stuff from the Makefile

### Checklist

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -1,20 +1,38 @@
 name: Pipeline
 
 on:
-  push:
-    branches: [main]
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [main, next]
 
 jobs:
-  CI:
-    name: CI
+  linux:
     strategy:
       fail-fast: false
-      matrix:
-        os: [ubuntu-latest, macos-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Docker
+        uses: docker/setup-buildx-action@v1
+
+      - name: Download deps
+        run: docker build . -t ci
+
+      # Disable network so that Azure specific behavior doesn't happen
+      - name: Quality
+        run: docker run --network none --env GOPATH="" --rm -v $PWD:/go/work ci make quality
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/pipeline.yml
```diff
@@ -1,20 +1,38 @@
 name: Pipeline
 
 on:
-  push:
-    branches: [main]
```
werne2j, line 5: Why not run this on a push to main?

.github/workflows/pipeline.yml
```diff
@@ -1,20 +1,38 @@
...
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [main, next]
 
 jobs:
-  CI:
-    name: CI
+  linux:
```
werne2j, line 9: Interesting to split them up by OS. Could we try adding windows back?

.github/workflows/pipeline.yml
jkayani: We can but I didn't see the point - we know the tests will pass and how the coverage will look from the results of the PR that the push to main came from, right? I can put it back just didn't really see the point

.github/workflows/pipeline.yml
jkayani: Sure, though I don't see the value of that unless we start publishing Windows binaries, which we could certainly do.  But IIRC we had problems with the tests on Windows a long time ago. Also the Azure-specific behavior applies to workloads in Windows and Linux VMs so we'd have to run in Docker on Windows which would make things even slower

werne2j: Yeah, you're right, we are not doing windows binaries so no need

werne2j (approved):

Issue #194, opened 2021-08-31
Add versioning to the docs


werne2j:
As we start to add more features and release new versions, we should version the docs so that users can find docs for the particular version they are using. 

Couple different ways to do it:
mkdocs with mike - https://github.com/squidfunk/mkdocs-material-example-versioning
mkdocs with readthedocs - https://docs.readthedocs.io/en/stable/versions.html

jkayani:
Works and includes a banner message if someone is browsing the `main` docs: https://ibm.github.io/argocd-vault-plugin/main/

PR #202, opened 2021-09-09
ci: Versioned docs via mike

jkayani:
### Description

Version docs via `mike`. Creates 2 moving aliases: `latest`, for the main branch of docs, and `stable` for the latest tagged release. Versions are published and aliases are updated on tag push/main push. Aliases are used off the GH pages URL to redirect to the version they point to.

Live example: https://jkayani.github.io/argocd-vault-plugin/

**Fixes:** #194 

### Checklist

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -3,6 +3,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
 jobs:
   deploy:
     runs-on: ubuntu-latest
@@ -11,5 +13,15 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: 3.x
-      - run: pip install mkdocs-material 
-      - run: mkdocs gh-deploy --force
+      - name: Setup deps
+        run: |
+          pip install mkdocs-material 
+          pip install mike
+          sudo apt-get install -y pcregrep
+      - name: Publish docs
+        run: | 
+          export REF=$(echo $GITHUB_REF | pcregrep --only-matching=1 ".*\/.*\/(.*)")
+          git config user.name mkdocs
+          git config user.email noreply@null.com 
+          git fetch origin gh-pages --depth=1
+          mike deploy -u -p --rebase $REF $(if [ $REF == "main" ]; then echo preview; else echo stable; fi)
--- a/docs/_overrides/main.html
+++ b/docs/_overrides/main.html
@@ -1,0 +1,23 @@
+{% extends "base.html" %}
+
+
+{% block announce %}
+  <div id="version-banner">
+  <script>
+    if (window.location.pathname.match("^\/argocd-vault-plugin\/main\/.*") !== null) {
+      const banner = document.createElement("h4");
+      const msg = document.createTextNode("These docs are from the main branch and may reference unreleased features. ");
+      const link = document.createElement("a");
+      link.setAttribute("style", "color: white;text-decoration: underline;");
+      link.setAttribute("href", `${window.location.origin}/argocd-vault-plugin/`)
+
+      link.appendChild(document.createTextNode("Go here for the latest released docs"))
+      banner.appendChild(msg);
+      banner.appendChild(link);
+
+      const currentDiv = document.getElementById("version-banner");
+      currentDiv.appendChild(banner);
+    }
+  </script>
+  </div>
+{% endblock %}
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -3,11 +3,16 @@ repo_url: https://github.com/IBM/argocd-vault-plugin
 
 theme:
   name: material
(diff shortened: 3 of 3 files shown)


.github/workflows/mkdocs.yml
```diff
@@ -11,5 +13,10 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: 3.x
-      - run: pip install mkdocs-material 
-      - run: mkdocs gh-deploy --force
+      - run: |
+          pip install mkdocs-material 
+          pip install mike
+      - run: | 
+        export REF=$(echo $GITHUB_REF | pcregrep --only-matching=1 ".*\/.*\/(.*)")
```
jkayani, line 20: This shell snippet works great locally (setting `GITHUB_REF` to stuff like `refs/heads/main` and `refs/tags/v1.2.3`). Need to be sure if this works in Actions though

jkayani: Tested here: https://github.com/jkayani/argocd-vault-plugin/runs/3561014531?check_suite_focus=true

jkayani:
The only complication is how to handle merging this, since we want docs based off v1.3.1 as the default (`stable`). But those can't be generated via CI since v1.3.1 was released already.

So we just have to do it manually:

- Apply the patch to `mkdocs.yaml` so that the version dropdown appears when generating the docs

- Run:
```
mike delete --all # to delete the existing HTML in gh-pages branch since we don't need it anymore
git checkout v1.3.1
mike deploy -u v1.3.1 stable

mike set-default stable
git push --force origin gh-pages:gh-pages
```

so that we add a v1.3.1 (alias `stable`) which is the default version of the docs to `gh-pages` branch (example commit: https://github.com/jkayani/argocd-vault-plugin/commit/ee051fad4c3e5d2efeb93de4844ecfd246c1880a)

- Then, merge this PR. It will run the Mkdocs action which will push a commit to `gh-pages` adding a `main` version (alias `latest`) on top. (example CI run: https://github.com/jkayani/argocd-vault-plugin/actions/runs/1219046108)

Result is versioned docs as if we had them the whole time: https://jkayani.github.io/argocd-vault-plugin. The GH pages URL goes to the default version; appending an alias redirects you to whatever version the alias points to

jkayani:
Proof it works on tags:

CI run: https://github.com/jkayani/argocd-vault-plugin/runs/3561655188?check_suite_focus=true

End result: https://jkayani.github.io/argocd-vault-plugin now points to v1.4.0 docs

werne2j:
@jkayani we will have to manually push each of the tags as well as 1.3.1, right?

werne2j (approved):
Looks good!

jkayani:
> @jkayani we will have to manually push each of the tags as well as 1.3.1, right?

If we wanted to retroactively release versioned docs, yes. But I don't know if that's really worthwhile - I think having the current release people are intended to use + the "preview" docs are better

> So why even do latest. Let’s just build the docs on a tag

I think it's worth having so a.) we can proof-read docs/make sure Mkdocs formats stuff right and b.) point people who need unreleased but available feature X to the docs

<hr/>

With all the above I think it might better to change things slightly: 

- alias the latest tagged release as `latest` (or just omit the alias, since we can just set the latest tagged version vx.y.yz as the default)
- alias the `main` based docs as `preview` 

That way it's more clear which is which. Thoughts?

werne2j:
> If we wanted to retroactively release docs for previous versions, yes. But I don't know if that's really worthwhile - I think having the current release people are intended to use + the "preview" docs is enough

I would like to be it retroactively, at least for the tags that have the docs.

I am fine with the other suggestions

jkayani:
Going to try to add a "Your looking at unreleased docs" on the preview version, and write a script to generate versioned docs for our previous releases (where applicable)

werne2j (approved):

jkayani:
- Updated PR to include a banner saying the `main` docs are for unreleased features and a link to the GH Pages URL which is set to be default, example here: https://jkayani.github.io/argocd-vault-plugin/main/
- Kept the `stable` alias so that even when we release new tags, the `stable` alias points to the latest tag and we only have to run `mike set-default` once when setting things up

jkayani:
Before merging this PR I will run:

```
mike delete --all
for tag that has mkdocs:
  git checkout $tag
  git cherry-pick versioned-docs
  mike deploy -u $tag stable
# manually inspect to make sure generated HTML looks right
mike set-default stable
git push --force origin gh-pages:gh-pages
```

And then this PR will create our `preview` docs and automate our Mkdocs workflow to do the above on tagged releases/merges to main

jkayani:
Versioned docs are pushed: https://github.com/IBM/argocd-vault-plugin/commits/gh-pages. This PR is ready to merge I think

werne2j (approved):

Issue #214, opened 2021-10-14
Add a new modifier to process non-string replacement data


teejaded:
**Is your feature request related to a problem? Please describe.**
I have significant amount data that is stored in json, in Vault.  I don't think it is currently possible to isolate specific sub-values in this case.

Consider the following test secret:
![Screen Shot 2021-10-14 at 4 37 44 PM](https://user-images.githubusercontent.com/3037973/137409668-fb24f8e7-395b-4c9f-8da5-7e8c71ecabc7.png)

If I need to isolate value of `.credentials.user` there's no way to do this currently.

**Describe the solution you'd like**

Add a new modifier that will allow us to select sub-sets of data possibly via jsonpath.

**Describe alternatives you've considered**
Duplicating data inside vault for use only with AVP.  This, for me, is sub-optimal since it would require additional automation or a manual process.

teejaded:
Okay, updated.
In the updated PR to add a modifier you just write a new function with the correct signature and add it into the map.

```golang
var modifiers = map[string]func([]string, interface{}) error{
	"base64encode": base64encode,
	"jsonPath":     jsonPath,
}
```

Now we can easily add new modifiers and pipe them together.

```
./bin/test-avp-path.sh '<path:secret/devops/test#credentials | jsonPath .pass | base64encode>' 'avp.kubernetes.io/kv-version: "1"'
apiVersion: v1
kind: Secret
metadata:
  annotations:
    avp.kubernetes.io/kv-version: "1"
  name: test
stringData:
  testpath: bXlwYXNz
---
```

Wrote a quick `fromJSON` for fun.  I'm not sure if it's actually useful so I didn't commit it.

```
./bin/test-avp-path.sh '<path:secret/devops/test#embeddedJson>' 'avp.kubernetes.io/kv-version: "1"'
apiVersion: v1
kind: Secret
metadata:
  annotations:
    avp.kubernetes.io/kv-version: "1"
  name: test
stringData:
  testpath: '{"key": "value"}'
  
./bin/test-avp-path.sh '<path:secret/devops/test#embeddedJson | fromJSON | jsonPath .key>' 'avp.kubernetes.io/kv-version: "1"'
apiVersion: v1
kind: Secret
metadata:
  annotations:
    avp.kubernetes.io/kv-version: "1"
  name: test
stringData:
  testpath: value
---
```

teejaded:
Feature merged! 

Docs here:
[...]

PR #215, opened 2021-10-14
feat: jsonPath modifier

teejaded:
### Description
adds jsonPath modifier for processing non-string replacement values

**Fixes:** https://github.com/IBM/argocd-vault-plugin/issues/214

### Checklist

### Type of Change

- [ ] Bugfix
- [x] Feature
- [x] New tests
- [ ] Other (please describe)

### DIFF
--- a/docs/howitworks.md
+++ b/docs/howitworks.md
@@ -265,7 +265,7 @@ This can be used for both generic and inline-path placeholders.
 
 ##### `jsonPath`
 
-The jsonPath modifier allows you use json path to post-process objects retrieved from a secrets manager before injecting into a Kubernetes secret.  The output is always a string.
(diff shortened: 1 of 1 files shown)


werne2j:
@teejaded Thanks for the PR, based on the issue, i want to make sure this will work with the `base64encode` modifier

werne2j:
@teejaded doing a little testing and noticed that this does not work with IBM or GCP. 

I see `adds jsonPath modifier for processing non-string replacement values` in the PR and am wondering if this should also work for strings? For example in GCP/IBM i added `{"key": "value"}` as the value of a secret. When we get that value from IBM/GCP it is a string, but i would think i should be able to use `jsonPath` to pull `.key` out. 

What i am seeing is `key is not found`.

PR #231, opened 2021-11-10
fix: Error when all manifests are empty

jkayani:
### Description

Until this PR, AVP will not error if all-empty manifests are given in the `generate` command. This means we get the below behavior:
```
$ ((>&2 echo "some error" && exit 1) | ./argocd-vault-plugin generate -c ~/Documents/ops/argo/vault/ibm-sm-creds.yaml -); echo $?;

some error
0
```
and this can be problematic when the `echo "some error"` is actually something like `helm template` failing when using AVP+Helm. The result will be that Argo CD believes all manifests for the given Argo app should be pruned with no obvious explanation why, and no errors in the Argo UI.

With this PR the behavior changes so that AVP prints a helpful error message:
```
$ ((>&2 echo "some error" && exit 1) | ./argocd-vault-plugin generate -c ~/Documents/ops/argo/vault/ibm-sm-creds.yaml -); echo $?;
some error
Error: No manifests
Usage:
  argocd-vault-plugin generate <path> [flags]

Flags:
  -c, --config-path string   path to a file containing Vault configuration (YAML, JSON, envfile) to use
  -h, --help                 help for generate
  -s, --secret-name string   name of a Kubernetes Secret containing Vault configuration data in the argocd namespace of your ArgoCD host (Only available when used in ArgoCD)

1
```

That way a user knows that something is wrong with the registered plugin invocation even if they cannot see the `stderr` of the previous command of the pipeline in the Argo UI.

### Checklist

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)


cmd/generate_test.go
```diff
@@ -32,6 +32,7 @@ func TestMain(t *testing.T) {
 		c := bytes.NewBufferString("")
 		cmd.SetArgs(args)
 		cmd.SetErr(c)
+		cmd.SetOut(bytes.NewBufferString(""))
```
jkayani, line 35: Did this all in the error-path generate tests to cut down on the useless output (the "Usage" instructions Cobra prints by default when there is an error)

werne2j (approved):
Looks great!

Issue #232, opened 2021-11-12
Extra modifiers


denraf:
**Is your feature request related to a problem? Please describe.**
We upload a base64 encoded yaml file into our vault. When retrieving this file with the plugin, we need to be able to base64 decode it and parse it so we can get a specific value out of the yaml file.

**Describe the solution you'd like**
2 extra modifiers:
- base64decode
- yamltojson

werne2j:
Those will be great additions, thanks @denraf!

werne2j:
These modifiers were released in https://github.com/IBM/argocd-vault-plugin/releases/tag/v1.6.0

PR #233, opened 2021-11-12
Denraf/feature/base64decode

denraf:
### Description
Adding extra modifier: base64decode

**Fixes:** https://github.com/IBM/argocd-vault-plugin/issues/232

### Checklist
- [X] Reviewed the guidelines for contributing to this repository
- [X] Tests for the changes have been updated

### Type of Change

- [ ] Bugfix
- [X] Feature
- [ ] New tests
- [ ] Other (please describe)

### Other information
I have a build with these two plugins running in my test environment.
Full path to secret:
`secret: <path:path/to/secrets#encoded_secret| base64decode  | jsonPath {.secret_key} >`

### DIFF
--- a/docs/howitworks.md
+++ b/docs/howitworks.md
@@ -271,9 +271,9 @@ Valid examples:
 
 - `<b64_username | base64decode>`
 
-- `<path:secrets/data/my-db#b64_username | base64encode>`
+- `<path:secrets/data/my-db#b64_username | base64decode>`
 
-- `<path:secrets/data/my-db#b64_username#version3 | base64encode>`
+- `<path:secrets/data/my-db#b64_username#version3 | base64decode>`
 
 ##### `jsonPath`
 
--- a/docs/howitworks.md
+++ b/docs/howitworks.md
@@ -263,6 +263,18 @@ Valid examples:
 
 This can be used for both generic and inline-path placeholders.
 
+##### `base64decode`
+
+The base64decode modifier decodes base64 encoded values into plain-text.
+
+Valid examples:
+
+- `<b64_username | base64decode>`
+
+- `<path:secrets/data/my-db#b64_username | base64encode>`
+
+- `<path:secrets/data/my-db#b64_username#version3 | base64encode>`
+
 ##### `jsonPath`
 
 The jsonPath modifier allows you use jsonpath to post-process objects or json, retrieved from a secrets manager, before injecting into a Kubernetes manifest.  The output is a string.  If your desired datatype is not a string, pass the output through jsonParse.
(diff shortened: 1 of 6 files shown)


werne2j:
Hi @denraf thanks for this contribution! Could you please do me a favor and split the two modifiers into separate pull requests? This will help us review and merge them independently. Thanks again!

jkayani (commented):
This looks good @denraf - would you mind adding some e2e tests to the `fixtures/` directory? Specifically creating a YAML that uses these modifiers both individually and separately in `fixtures/input/nonempty/` and then adding the resulting output to `fixtures/output/all.yaml`?

denraf:
I've been looking into the e2e tests and was wondering where the data comes from that's being used in the output. Any pointers?

werne2j:
Hi @denraf for e2e tests.. this is where the e2e tests are https://github.com/IBM/argocd-vault-plugin/blob/main/cmd/generate_test.go#L113 and this is the Test Vault we execute the command against https://github.com/IBM/argocd-vault-plugin/blob/a270631026d83a9167cd13c9e72a100e129f3556/pkg/helpers/test_helpers.go#L109 so you should be able to see where we are writing to this test vault where we then reference in yaml files in the `fixtures/input/nonempty/` directory. Hopefully that helps!

denraf:
That was exactly what I was looking for. Thank you for you help!

werne2j:
@denraf Could you do me a favor and squash your commits down to two. One for `base64decode` and one for `yamltojson`?

teejaded:
Since modifiers can pass objects through the pipeline would it make sense to split yamltojson up into separate modifiers?

* fromYaml
* toYaml
* toJson

werne2j:
@teejaded I think that is an interesting idea and seems to make sense. You are right, as we can string together multiple modifiers. @denraf Let's change this PR to be just `base64decode` and then we can work on the other modifiers we want to include in other PRs.

denraf:
@teejaded @werne2j I'm not sure it makes sense to split the modifiers up in the suggested way. The standard for anything vault related is json, so when transforming data, it should end up in json.

I will split the PR up for `base64decode` and the `yamltojson` modifier.

werne2j:
@denraf Can you add documentation to the modifier here https://github.com/IBM/argocd-vault-plugin/blob/main/docs/howitworks.md#modifiers following the same pattern as the other modifiers

werne2j:
Looks like the e2e tests are failing - https://github.com/IBM/argocd-vault-plugin/runs/4301432074?check_suite_focus=true#step:5:11

werne2j (approved):
Looks good!

docs/howitworks.md
```diff
@@ -263,6 +263,18 @@ Valid examples:
...
 
+##### `base64decode`
+
+The base64decode modifier decodes base64 encoded values into plain-text.
+
+Valid examples:
+
+- `<b64_username | base64decode>`
+
+- `<path:secrets/data/my-db#b64_username | base64encode>`
```
jkayani, line 274: I think you mean `decode` here :)

jkayani (approved):

docs/howitworks.md
jkayani: Feel free to fix here, code looks great otherwise so I left an approval

denraf: Great catch! Just updated.

PR #249, opened 2021-12-29
chore: Update to go 1.17 and update deps

jkayani:
### Description

Updates deps and updates to using go 1.17. Some special things to call out:

- IBM SM SDK removed the `CertificateSecretData` struct they were using to return cert data for `public_cert` and `imported_cert` secret types; it has been replaced with a `map[string]interface{}` which simplifies our code. Code and tests updated, and confirmed to work in an e2e test. 

- Google `gnostic` library was moved across repos which go doesn't seem to understand yet. Fixed with the suggestion given here https://github.com/google/gnostic/issues/262

- The Vault auth libraries for `approle` and `userpass` have a weird revision that Go doesn't seem to understand. It is fixed in `main` but not any of the stable released tags. https://github.com/hashicorp/vault/issues/13528

- Should address `CVE-2021-43998` for Vault (though I don't think we exposed the vulnerable code here) and seems to remove the `ulikuntiz/xz` dep to handle `CVE-2020-16845`

### Checklist

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)


go.mod
```diff
@@ -1,55 +1,213 @@
...
-	k8s.io/klog/v2 v2.20.0 // indirect
-	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b // indirect
+	k8s.io/apimachinery v0.23.1
+	k8s.io/client-go v0.23.1
+	k8s.io/klog/v2 v2.40.1 // indirect
+	k8s.io/utils v0.0.0-20211208161948-7d6a63dca704 // indirect
 	sigs.k8s.io/yaml v1.3.0
 )
+
+require (
```
jkayani, line 59: I guess the go 1.17 changes the format of this file to include 2 `require()` blocks, not sure what that's about yet

Dockerfile
```diff
@@ -1,13 +1,15 @@
...
+RUN apk add make git
 
 ADD go.mod go.mod
 ADD go.sum go.sum
 
 ENV GOPATH=""
 ENV CGO_ENABLED=0
-RUN go mod download
+
+RUN go mod tidy -compat=1.17
```
jkayani, line 12: I'm replacing `go mod download` with this because the former causes this error since upgrading to go 1.17: 
```
go mod tidy: error loading go 1.16 module graph: github.com/hashicorp/vault@v1.9.2 requires
        github.com/hashicorp/vault/api/auth/approle@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```
which I can reproduce when running `go mod tidy` without the `-compat=1.17` flag. As it turns out `go mod tidy` will not only delete unneeded dependencies, but download the ones strictly needed for the main package according to https://github.com/golang/go/issues/43994#issuecomment-770053099. So it seems that is enough to get what we want.

go.mod
jkayani: > Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies.

https://go.dev/doc/go1.17#tools

werne2j (approved):
Looks good!

PR #265, opened 2022-01-13
feat(backend): add Sops Backend

Icerius:
### Description
Add support for secrets encrypted with SOPS and stored in a local file.
The actual backend is currently called LocalSecretManagerBackend since it has no dependancy on sops, i.e. the decrypt function is passed as a parameter when creating the backend.

This first iteration contains the bare minimum for a working version, i.e. only json and yaml (both parsed as yaml) and only simple keys.

**Fixes:** <! -- link to issue -->

### Checklist

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] New tests
- [ ] Other (please describe)


jkayani (commented):
Hi @Icerius thanks for the substantial work here! A few questions (sorry if they're elementary, I'm not familiar with `sops`):

- How does `sops` know how to decrypt the file? Presumably it's using a decryption key set somewhere? Assuming that's the case, we should allow this key to be configured as an env var in the repo-server container, via config file, or from a key in a Kubernetes secret like we do for all the backend specific config keys like `VAULT_TOKEN`, etc: https://github.com/argoproj-labs/argocd-vault-plugin/blob/main/pkg/config/config.go#L203

- > This first iteration contains the bare minimum for a working version ... and only simple keys.

Can you elaborate on what this means? What kinds of keys can a user expect to work, and which ones wouldn't? This would be good information for the docs section

Icerius:
> * How does `sops` know how to decrypt the file? Presumably it's using a decryption key set somewhere? Assuming that's the case, we should allow this key to be configured as an env var in the repo-server container, via config file, or from a key in a Kubernetes secret like we do for all the backend specific config keys like `VAULT_TOKEN`, etc: https://github.com/argoproj-labs/argocd-vault-plugin/blob/main/pkg/config/config.go#L203

For key material sops supports AWS KMS, GCP KMS, Azure, Vault, GPG and AGE.
A encrypted file contains metadata on what keys it was encrypted with.
When using GPG or AGE keys there is a standard lookup path which can be changed using an environment variable.
When using managed services AWS, GCP, etc. the credentials can be handled similar to how credentials are handled for current backends.
More detail can be found here: https://github.com/mozilla/sops
My current usage of SOPS is using AGE keys which I volume mount from a kubernetes secret to the standard sops location in the ArgoCD container.
I will need to look a bit more into if the credentials for the other services needs some extra handling or can piggy back on current usage.
  
> * > This first iteration contains the bare minimum for a working version ... and only simple keys.
> 
> Can you elaborate on what this means? What kinds of keys can a user expect to work, and which ones wouldn't? This would be good information for the docs section

To elaborate simple means only the first level keys in a document
With the following yaml file:
`
hello: Welcome to SOPS! Edit this file as you please!
parent:
    child: data
`
then path would be the location of the file and secret-key could be either hello or parent.

The reason I added this comment was that I originally thought that then there would not be a way to get the value of parent.child but after reading the documentation better I found that the product has more power then I realised with the support of jsonpath.

I plan to update the docs before this is completed but I wanted to get some feedback on the change itself first.

jkayani:
RE: configuration for sops decrypt key location, I see this https://github.com/mozilla/sops#22encrypting-using-age:

> When decrypting a file with the corresponding identity, sops will look for a text file name keys.txt located in a sops subdirectory of your user configuration directory. On Linux, this would be $XDG_CONFIG_HOME/sops/keys.txt. On macOS, this would be $HOME/Library/Application Support/sops/keys.txt. On Windows, this would be %AppData%\sops\keys.txt. You can specify the location of this file manually by setting the environment variable SOPS_AGE_KEY_FILE.

So what I'd like to see added to this PR is code to take any `AVP_SOPS_*` keys (read from env, config file, or secret) and set them in the environment. That way users can expect setting the above env variable through any of the AVP supported configs will be passed through to the sops tool. You can do this by just adding `sops` to this slice here: https://github.com/argoproj-labs/argocd-vault-plugin/blob/main/pkg/config/config.go#L41.

Icerius:
> So what I'd like to see added to this PR is code to take any AVP_SOPS_* keys (read from env, config file, or secret) and set them in the environment. That way users can expect setting the above env variable through any of the AVP supported configs will be passed through to the sops tool. You can do this by just adding sops to this slice here: https://github.com/argoproj-labs/argocd-vault-plugin/blob/main/pkg/config/config.go#L41.

Thanks for the info, I will look into making that change

jkayani (commented):
Aside from my comment about exposing all sops config options in AVP, this looks good to me (code, tests, docs). One small nit, other backend files are in all lowercase - can we change `backends/localSecretManager.go` to `backends/localsecretmanager.go`? 

I'm alright with keeping the name `localsecretmanager.go` since like you said, this is generic enough where it could be used to support something like an encrypted Ansible Vault. 

Ofc I'd like @werne2j to look at this in case he disagrees with any of my comments here.

Icerius:
> Aside from my comment about exposing all sops config options in AVP, this looks good to me (code, tests, docs). One small nit, other backend files are in all lowercase - can we change `backends/localSecretManager.go` to `backends/localsecretmanager.go`?

Done
 
>  You can do this by just adding sops to this slice

Nice feature, sops has been added to the slice

go.mod
```diff
@@ -198,6 +204,8 @@ require (
 	github.com/tv42/httpunix v0.0.0-20191220191345-2ba4b9c3382c // indirect
 	github.com/vmware/govmomi v0.18.0 // indirect
 	go.etcd.io/bbolt v1.3.6 // indirect
+	go.mozilla.org/gopgagent v0.0.0-20170926210634-4d7ea76ff71a // indirect
+	go.mozilla.org/sops/v3 v3.7.1 // indirect
```
jkayani, line 208: Just a small question, any idea why this shows up as an `indirect` dependency? I see the config code directly imports this module so I'd expect it to appear like the Vault libs, cloud SDKs, etc. appear in the go.mod list. Not a big deal necessarily but I'd like to be sure about it so we don't get weird behavior with dependency updates, etc.

jkayani (approved):
This all LGTM, thanks again for all the great work! Pending @werne2j 's approval as well.

go.mod
Icerius: I have been looking into this and `go mod tidy` does not appear to remove indirect without additional work.

Tidy currently reports few instances of

> but go 1.16 would select

If I run the recommended tidy command to fix it then indirect is removed from sops but that also changes many other dependancies and should therefore be a different PR

jkayani: I checked out your fork (`-b sops-backend`) and ran `go mod tidy -compat=1.17` and got this output:

```
Joshuas-MacBook-Pro-2:argocd-vault-plugin jkayani$ git diff go.mod 
diff --git a/go.mod b/go.mod
index 611e6ad..a291f07 100644
--- a/go.mod
+++ b/go.mod
@@ -43,6 +43,7 @@ require (
        github.com/spf13/cobra v1.3.0
        github.com/spf13/viper v1.10.1
        go.mongodb.org/mongo-driver v1.8.1 // indirect
+       go.mozilla.org/sops/v3 v3.7.1
        golang.org/x/crypto v0.0.0-20211108221036-ceb1ce70b4fa // indirect
        golang.org/x/net v0.0.0-20211216030914-fe4d6282115f
        golang.org/x/oauth2 v0.0.0-20211104180415-d3ed0bb246c8 // indirect
@@ -205,7 +206,6 @@ require (
        github.com/vmware/govmomi v0.18.0 // indirect
        go.etcd.io/bbolt v1.3.6 // indirect
        go.mozilla.org/gopgagent v0.0.0-20170926210634-4d7ea76ff71a // indirect
-       go.mozilla.org/sops/v3 v3.7.1 // indirect
        go.opencensus.io v0.23.0 // indirect
        go.uber.org/atomic v1.9.0 // indirect
        golang.org/x/sync v0.0.0-20210220032951-036812b2e83c // indirect
```

I think this is all I'd like to see - can you confirm this works the same way for you (it should, right?) and commit/push this if so?

Icerius: Did you also look at the change this makes to `go.sum`?

If you are comfortable with it then I will make this change

jkayani: I did - I'm not too concerned about them since none of the versions in `go.mod` are changing. Thanks for checking though

Icerius: The code has been updated with this change

jkayani:
Tested locally:

`sops-secret.yaml`:

```
secret: passw0rd
```

age keypair:

```
AGE-SECRET-KEY-???
age1404kpjnjw4hf2zma5hljqakkcmkmgu0yewn2khcgzdhj5wln5v6qwz2rkt
```

encrypted sops-secret.yaml to yield sops-secret-enc.yaml:

```
secret: ENC[AES256_GCM,data:5WMJgXG8LQI=,iv:Itk8noU579Vqor68cVSuiL0KT4c+GOTHIRCvMcAk10o=,tag:K2RmQvoyevIhX/u6c2bBkQ==,type:str]
sops:
    kms: []
    gcp_kms: []
    azure_kv: []
    hc_vault: []
    age:
        - recipient: age1404kpjnjw4hf2zma5hljqakkcmkmgu0yewn2khcgzdhj5wln5v6qwz2rkt
          enc: |
            -----BEGIN AGE ENCRYPTED FILE-----
            YWdlLWVuY3J5cHRpb24ub3JnL3YxCi0+IFgyNTUxOSBjNDJWdTkwRGEzMDhMaTJq
            T29BRTB4bE5sNmR5TlBrQ1JWZ3BnZFoxdXpNClFwbDJHRENVcWV4OHYvRkpzc25r
            aG8xUTNvN0ZiMmJ2V0VVYTY0aG1jVDAKLS0tIE5zbmxIWElhQ1pGbS83dUJ6eXpL
            V3BXQWY4c2VETTM3V2hmaUhLUVFxRzAKufN2bEdwlnx2jHreBwrQf5z47TlrLz/9
            h9SoaxuYRNcE7uEv9ejMVg5csHAwo3GPdno/6UXvTioVH1P0VMWzZQ==
            -----END AGE ENCRYPTED FILE-----
    lastmodified: "2022-01-18T16:18:14Z"
    mac: ENC[AES256_GCM,data:bRtow8CvkVEFDkFSXsJitDZqUkADIa6TiSKlRtVZmxvoBS6hkrYHRkRKy5/ldflymg2trUA02mUTslzqyB++LHu7TjVIDQCUXd4V88E1COatbcMjXSMfIYwmHlrmvuJTa2DN2j6svBL4sVhln1DgCCl/pfUOv6BfL/ze0e4kiGw=,iv:Wr+W+R98D7dcGHC2/gER8cFoCcDThItO6vaScetiAtU=,tag:zxpsKFbEhhaw8yQaA/3jrQ==,type:str]
    pgp: []
    unencrypted_suffix: _unencrypted
    version: 3.7.1
```

Created k8s secret manifest:

```
apiVersion: v1
kind: Secret
metadata:
  annotations:
    avp.kubernetes.io/path: "fixtures/input/local/sops-secret-enc.yaml"
    avp.kubernetes.io/remove-missing: "true"
    # avp.kubernetes.io/secret-version: "1"
    # avp.kubernetes.io/kv-version: "1"
  name: example-secret
  namespace: default
type: Opaque
data:
  secret: <secret>
```

Generate manifest:
```
Joshuas-MacBook-Pro-2:argocd-vault-plugin jkayani$ SOPS_AGE_KEY_FILE=$PWD/fixtures/input/local/sops-age-key-file AVP_TYPE=sops ./argocd-vault-plugin generate fixtures/input/local/sops.yaml
apiVersion: v1
data:
  secret: passw0rd
kind: Secret
metadata:
  annotations:
    avp.kubernetes.io/path: fixtures/input/local/sops-secret-enc.yaml
    avp.kubernetes.io/remove-missing: "true"
  name: example-secret
  namespace: default
type: Opaque
---
```

go.mod
```diff
@@ -60,7 +60,10 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require go.mozilla.org/sops/v3 v3.7.1
```
werne2j, line 63: I would like to see this put in with the rest of the requirements.

Icerius: I do not know why tidy decided to keep it standalone.
Will move and see if tidy is still happy

Icerius: This has now been fixed

werne2j:
@Icerius Could you resolve the conflicts and then we will finish our review. Thanks!

Icerius:
> @Icerius Could you resolve the conflicts and then we will finish our review. Thanks!

Done

werne2j (approved):
Looks good! Thanks @Icerius!

PR #281, opened 2022-01-23
feat: support for go templates from an annotation

hugodutka:
### Description
Introduce support for specifying a go template inside an annotation to generate and replace the contents of a manifest's data field.

Please look at `docs/howitworks.md` to see how the feature works exactly.

My goal was to enable fully dynamic generation of secrets' contents. My problem was that I had to repeat myself every time I created a secret. First I had to create the secret in the vault and specify all the fields it should have. Then in the manifest I used the same field names for clarity. Then for every secret key in the manifest I had to specify the secret's path. Templates now allow me to specify all the fields in the Vault once and have the manifest be generated automatically.

This feature works similarly to [Vault's agent injector](https://www.vaultproject.io/docs/platform/k8s/injector).

### Checklist

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] New tests
- [ ] Other (please describe)

### Other information
I went ahead and implemented the functionality but did not talk to the maintainers to see if you'd be interested in introducing it. Would you like to add it or do you have different plans for this plugin?

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -68,3 +68,4 @@ We support several different annotations that can be used inside a kubernetes re
 | avp.kubernetes.io/kv-version | Version of the KV Secret Engine |
 | avp.kubernetes.io/secret-version | Version of the secret to retrieve. Only effective on generic `<placeholder>`s so `avp.kubernetes.io/path` is required when this annotation is used |
 | avp.kubernetes.io/remove-missing | Plugin will not throw error when a key is missing from Vault Secret. Only works on `Secret` or `ConfigMap` resources |
+| avp.kubernetes.io/data-template | A Go template used to generate and replace the contents of the manifest's data field |
--- a/docs/howitworks.md
+++ b/docs/howitworks.md
@@ -246,6 +246,53 @@ stringData:
 ```
 This only works with _generic_ placeholders.
 
+##### Data templates
+If the `avp.kubernetes.io/data-template` annotation is present, the plugin will interpret its
+value as a [go template](https://pkg.go.dev/text/template) and use it to generate and replace
+the contents of the manifest's `data` field. This annotation should be used in tandem with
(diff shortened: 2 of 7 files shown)


jkayani:
The code looks great @hugodutka - I'm just a little confused on the use-case. I read the nice docs you wrote here, but if I want to generate a secret with all the keys of a particular Vault secret, why wouldn't I just use something like the agent injector, as I have no need to finely control where the keys of the secret get injected in the YAML via `<placeholder>`s?

Is the thinking here so that you can generate manifests that can be managed with Argo CD, unlike the agent injector which will just inject the secrets in pods? Or maybe that this would work with other backends besides Hashicorp Vault? I guess in theory it should, but has it been tested with any other?

werne2j:
Thanks again for the PR @hugodutka, after some discussions we decided not to include this feature. I think if we were going to do something like this, it would be to make Go templating a fully supported templating option that could be used instead of using the `< >` placeholders. That may be something we consider for a v2 of AVP but it will not be in our immediate backlog. Thanks again!

bdoublet91:
Hi,
I started with AVP and I find this use case very interesting.

The last thing about this is from https://github.com/argoproj-labs/argocd-vault-plugin/issues/305

> Thanks for the suggestion @roee-hershko! I agree with @jkayani here. We take a declarative approach, in the sense that if you want to add a new item to a secret, it should be done in git. This prevents the situation of not knowing what secrets are in Kubernetes when looking in git.

In my case, I use vault as full env file kv for my application. I setup the secret key ref to my secret file rendered via AVP in my chart deployment.

But I would list all keys on KV path and render the secret file to avoid a nested dict json workaround...
image

Because our devs doesn' t know k8s and also we limit interaction with our git helm, they only released app version and interact with vault UI to add env variable (secret or not) and argoCD + AVP do the deployment.

Could you consider to review your use case ?

Thanks for AVP, it rocks ;)

Issue #235, opened 2021-11-17
Any plan to support 1Password


Klaudioz:
Hello, do you have any plan to support [1Password Kubernetes Operator](https://github.com/1Password/onepassword-operator)?

werne2j:
Hi @Klaudioz It is not in our immediate backlog but we can take a look. We also accept contributions so if you or anyone else would like to contribute 1password as a backend we would love to review and work to get it into the plugin.

josephlim75:
If you are using vault, you can install 1password plugin so that vault can sync with 1password.  The you can use argocd vault to retrieve from 1password vault secret path

werne2j:
Support released in https://github.com/argoproj-labs/argocd-vault-plugin/releases/tag/v1.10.0

maxisam:
@werne2j Thanks for making this work. However, it is not base64 encoded when I pull it. Is it possible to encode it? or an option to do it? Thanks!

PR #303, opened 2022-02-22
feat: Support 1Password Connect as a backend

lucasmarshall:
### Description
Add support for 1Password Connect servers as a backend

**Fixes:** #235

### Checklist
- [X] Reviewed the guidelines for contributing to this repository
- [X] Tests for the changes have been updated
- [X] Docs have been added / updated

### Type of Change

- [ ] Bugfix
- [X] Feature
- [X] New tests
- [X] Documentation content changes
- [ ] Other (please describe)

### DIFF
--- a/docs/backends.md
+++ b/docs/backends.md
@@ -73,7 +73,7 @@ In order to use Kubernetes Authentication a couple of things are required.
           automountServiceAccountToken: true
     ```
 
-2. Configuring Kubernetes  
+2. Configuring Kubernetes
     Use the /config endpoint to configure Vault to talk to Kubernetes. Use `kubectl cluster-info` to validate the Kubernetes host address and TCP port. For the list of available configuration options, please see the [API documentation](https://www.vaultproject.io/api/auth/kubernetes).
 
     ```
@@ -159,7 +159,7 @@ data:
 **Note**: Only Vault KV-V2 backends support versioning. Versions specified with a KV-V1 Vault will be ignored and the latest version will be retrieved.
 
 ### IBM Cloud Secrets Manager
-For IBM Cloud Secret Manager we only support using IAM authentication at this time. 
+For IBM Cloud Secret Manager we only support using IAM authentication at this time.
 
 We support all types of secrets that can be retrieved from IBM Cloud Secret Manager. Please note:
 
@@ -511,4 +511,44 @@ data:
   current-password: <password>
   current-password-again: <path:secret-id#password#version-id>
   password-old: <path:secret-id#password#old-version-id>
+```
+
+### 1Password Connect
+##### 1Password Connect Authentication
+Refer to the [1Password Secrets Automation overview](https://support.1password.com/secrets-automation/) for 1Password Connect usage.
+
+These are the parameters for YCL:
+
+```
+AVP_TYPE: 1passwordconnect
+AVP_OP_CONNECT_TOKEN: Your 1Password Connect access token
+AVP_OP_CONNECT_HOST: The hostname of your 1Password Connect server
+```
+
+##### Examples
+
+###### Path Annotation
+
+```yaml
+kind: Secret
+apiVersion: v1
+metadata:
+  name: test-secret
+  annotations:
+    avp.kubernetes.io/path: "vaults/vault-uuid/items/item-uuid"
+type: Opaque
+data:
+  password: <key>
+```
+
+###### Inline Path
+
+```yaml
+kind: Secret
+apiVersion: v1
+metadata:
+  name: test-secret
+type: Opaque
+data:
+  password: <path:vaults/vault-uuid/items/item-uuid#key>
 ```
--- a/docs/backends.md
+++ b/docs/backends.md
@@ -521,8 +521,8 @@ These are the parameters for YCL:
 
 ```
 AVP_TYPE: 1passwordconnect
-AVP_OP_CONNECT_TOKEN: Your 1Password Connect access token
-AVP_OP_CONNECT_HOST: The hostname of your 1Password Connect server
+OP_CONNECT_TOKEN: Your 1Password Connect access token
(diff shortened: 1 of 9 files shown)


docs/backends.md
```diff
@@ -512,3 +512,43 @@ data:
...
+### 1Password Connect
+##### 1Password Connect Authentication
+Refer to the [1Password Secrets Automation overview](https://support.1password.com/secrets-automation/) for 1Password Connect usage.
+
+These are the parameters for YCL:
+
+```
+AVP_TYPE: 1passwordconnect
+AVP_OP_CONNECT_TOKEN: Your 1Password Connect access token
+AVP_OP_CONNECT_HOST: The hostname of your 1Password Connect server
```
werne2j, line 525: If `OP_CONNECT_TOKEN` and `OP_CONNECT_HOST` are env variables that 1password connect will read by default, https://github.com/1Password/connect-sdk-go#environment-variables, i would rather we strip the `AVP_` and just support the native env vars like we do with Vault and others.

This would also require adding `op_connect` to this list: https://github.com/argoproj-labs/argocd-vault-plugin/blob/main/pkg/config/config.go#L44

lucasmarshall: Sure, that's reasonable. Will push a new commit shortly

pkg/config/config_test.go
```diff
@@ -157,6 +157,14 @@ fDGt+yaf3RaZbVwHSVLzxiXGsu1WQJde3uJeNh5c6z+5
 			},
 			"*backends.LocalSecretManager",
 		},
+		{
```
werne2j, line 160: Can you also add a case for a using the one password backend with a missing variable in the TestNewConfigMissingParameter test

lucasmarshall: Sure

lucasmarshall: Just realized I neglected to push the additional test I wrote to the correct remote. Fixed now.

werne2j (approved):
Looks good, will let @jkayani take a look before merging

go.mod
```diff
@@ -55,6 +55,7 @@ require (
 require (
 	cloud.google.com/go v0.99.0 // indirect
 	filippo.io/age v1.0.0-beta7 // indirect
+	github.com/1Password/connect-sdk-go v1.2.0 // indirect
```
jkayani, line 58: This is a direct dependency isn't it? Wondering why it's marked as `// indirect` then

pkg/config/config.go
```diff
@@ -213,6 +214,24 @@ func New(v *viper.Viper, co *Options) (*Config, error) {
 
 			backend = backends.NewYandexCloudLockboxBackend(sdk.LockboxPayload().Payload())
 		}
+	case types.OnePasswordConnect:
+		{
+			if !v.IsSet(types.EnvOPConnectToken) ||
```
jkayani, line 219: Wondering if this is necessary, assuming that the `OP_CONNECT` vars are the ones expected by the SDK; omitting this check and deferring to the SDK means we can support any auth methods this backend supports going forward with no code changes, as long as we set all `OP_CONNECT` prefixed config keys in the environment (which your change to add `op_connect` to the backend prefixes slice does). Not a big deal, just a thought

docs/config.md
```diff
@@ -16,74 +18,88 @@ metadata:
...
+| AVP_YCL_SERVICE_ACCOUNT_ID | Yandex Cloud Lockbox service account ID             | Required with `TYPE` of `yandexcloudlockbox`                                                                                                                                 |
+| AVP_YCL_KEY_ID             | Yandex Cloud Lockbox service account Key ID         | Required with `TYPE` of `yandexcloudlockbox`                                                                                                                                 |
+| AVP_YCL_PRIVATE_KEY        | Yandex Cloud Lockbox service account private key    | Required with `TYPE` of `yandexcloudlockbox`                                                                                                                                 |
+| OP_CONNECT_TOKEN           | 1Password Connect access token                      | Required with `TYPE` of `1passwordconnect`                                                                                                                                   |
```
jkayani, line 76: If `OP_CONNECT` prefixed env vars are what the SDK expects I don't think we need to explicitly call them out here, the note above this table should point users to the right place:

> We support all the backend specific environment variables each backend's SDK will accept (e.g, `VAULT_NAMESPACE`, `AWS_REGION`, etc). Refer to the [specific backend's documentation](./backends) for details.

Otherwise, we should keep them here

docs/backends.md
```diff
@@ -512,3 +512,43 @@ data:
   current-password-again: <path:secret-id#password#version-id>
   password-old: <path:secret-id#password#old-version-id>
 ```
+
+### 1Password Connect
```
jkayani, line 516: Does this backend support versioning? I don't see anything to suggest it does in the code, in which case we should add an explicit note here that specified versions will be ignored

jkayani (commented):
Code, tests and docs look great! Thanks for the work @lucasmarshall. I'd like to resolve at least the dependency question since the OP SDK should be marked as a direct dependency (thinking about build issues if dependencies aren't categorized correctly). The other things aren't big deals but I leave it to you to make any changes if you see fit.

docs/backends.md
lucasmarshall: It does not. I will add a note

go.mod
```diff
@@ -52,6 +52,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/1Password/connect-sdk-go v1.2.0
```
werne2j, line 55: Sorry, one small thing. I would like this to be up in the above require block. Seems a little odd to have another require block with just one package.

werne2j (changes requested):

go.mod
lucasmarshall: Sure. `go mod tidy` put this here. Not sure why.

PR #324, opened 2022-03-22
docs: add note about command injection for helm example

werne2j:
### Description
Add a note to the helm docs about the dangers of malicious command injection for example

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -73,6 +73,8 @@ Helm args must be defined in the application manifest:
           value: -f values-dev.yaml -f values-dev-tag.yaml
 ```
 
+**Note: Bypassing the parameters like this can be dangerous in a multi-tenant environment as it could allow for malicous injection of arbitrary commands. So be cautious when doing something like in a production environment. Making sure that proper permissions and protections are very important when doing something liek this.** 
+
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -73,6 +73,8 @@ Helm args must be defined in the application manifest:
           value: -f values-dev.yaml -f values-dev-tag.yaml
 ```
 
+**Note: Bypassing the parameters like this can be dangerous in a multi-tenant environment as it could allow for malicous injection of arbitrary commands. So be cautious when doing something like in a production environment. Making sure that proper permissions and protections are very important when doing something liek this.**
```
todaywasawesome, line 76:
```suggestion
**Note: Bypassing the parameters like this can be dangerous in a multi-tenant environment as it could allow for malicious injection of arbitrary commands. So be cautious when doing something like in a production environment. Ensuring proper permissions and protections is very important when doing something like this.** 
```

crenshaw-dev, line 76:
```suggestion
**Note: Bypassing the parameters like this can be dangerous in a multi-tenant environment as it could allow for malicious injection of arbitrary commands. So be cautious when doing something like in a production environment. Making sure that proper permissions and protections are enforced is very important when doing something like this.** 
```

crenshaw-dev (changes requested):

jkayani (approved):
Thanks

PR #352, opened 2022-06-13
feat: Support AVP installed as CMP sidecar

jkayani:
### Description

**Fixes:** <! -- link to issue -->

### Checklist

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)


docs/usage.md
```diff
@@ -75,15 +131,45 @@ Helm args must be defined in the application manifest:
...
+  avp-helm.yaml: |
+    ---
+    apiVersion: argoproj.io/v1alpha1
+    kind: ConfigManagementPlugin
+    metadata:
+      name: argocd-vault-plugin-helm
+    spec:
+      allowConcurrency: true
+
+      # Note: this command is run _before_ any Helm templating is done, therefore the logic is to check
```
jkayani, line 156: Remove comment

docs/usage.md
```diff
@@ -128,6 +242,33 @@ configManagementPlugins: |
...
+  avp-jsonnet.yaml: |
+    ---
+    apiVersion: argoproj.io/v1alpha1
+    kind: ConfigManagementPlugin
+    metadata:
+      name: argocd-vault-plugin-kustomize
+    spec:
+      allowConcurrency: true
+
+      # Note: this command is run _before_ anything is done, therefore the logic is to check
```
jkayani, line 256: Remove comment

docs/usage.md
```diff
@@ -138,5 +279,14 @@ If you want to load in a new value from your Secret Manager without making any n
 
 You can also use the `argocd app diff` command passing the `--hard-refresh` flag. This will run argocd-vault-plugin again and pull in the new values from you Secret Manager and then you can either have Auto Sync setup or Sync manually to apply the new values.
 
-### Caching the Vault Token
+### Caveats
+
+#### Caching the Hashicorp Vault Token
 The plugin tries to cache the Vault token obtained from logging into Vault on the `argocd-repo-server`'s container's disk, at `~/.avp/config.json` for the duration of the token's lifetime. This of course requires that the container user is able to write to that path. Some environments, like Openshift, will force a random user for containers to run with; therefore this feature will not work, and the plugin will attempt to login to Vault on every run. This can be fixed by ensuring the `argocd-repo-server`'s container runs with the user `argocd`.
+
+#### Running argocd-vault-plugin in a sidecar container
```
jkayani, line 287: Mention the timeouts and how they may need to be adjusted if lots of placeholders: https://argo-cd.readthedocs.io/en/stable/user-guide/config-management-plugins/#using-a-cmp

jkayani (commented):
Reading over https://blog.argoproj.io/breaking-changes-in-argo-cd-2-4-29e3c2ac30c9 it looks like starting with 2.4.0, environment variables set at the application manifest get the `ARGOCD_ENV` prefix for `argod-cm` ConfigMap installed plugins, not just sidecars. So I guess existing docs need to be updated, and maybe the code should accept vars with or w/o that prefix (so `AVP_TYPE` or `ARGOCD_ENV_AVP_TYPE`)

docs/compatibility.md
```diff
@@ -12,3 +12,4 @@ If a version combination is missing from this table, **it is still very likely A
 | 1.20 (4.7) | 2.0.x | 1.x |
 | 1.21 (4.8) | 2.0.x | 1.x |
 | 1.21 (4.8) | 2.1.x | 1.x |
+| 1.23       | 2.4.0 | 1.x |
```
werne2j, line 15: This actually works with 2.4? Seems like it shouldn't based on the breaking changes.

jkayani: It does if you re-map the `ARGOCD_ENV` variables to their equivalents w/o that prefix, but I agree that it's best to remove this note and instead include it with #356

werne2j (approved):

timofey-drozhzhin:
@jkayani I noticed that the HELM plugin in [manifests/cmp-sidecar/cmp-plugin.yaml](https://github.com/argoproj-labs/argocd-vault-plugin/blob/08bfa36f6ed9f3838fcab484802df68f94fd0157/manifests/cmp-sidecar/cmp-plugin.yaml) requires both `Chart.yaml` and `values.yaml` files before generating the helm template. Wouldn't it make sense to include the `values.yaml` file in the generate command by default, since we already know it's there?

Before:
`helm template $ARGOCD_APP_NAME -n $ARGOCD_APP_NAMESPACE ${ARGOCD_ENV_HELM_ARGS} .`
After:
`helm template $ARGOCD_APP_NAME -n $ARGOCD_APP_NAMESPACE --values ./values.yaml ${ARGOCD_ENV_HELM_ARGS} .`

On a similar note, it might also be a good idea to limit the discover command to the app root (`-maxdepth 1`) to avoid false positives, since the generate command is also ran at the app directory level.

PS - I'm new here, and I understand the manifest is meant to be more of a starting point. So if anything, just ignore :)

timofey-drozhzhin:
@jkayani for [cmp-sidecar/cmp-plugin.yaml](https://github.com/argoproj-labs/argocd-vault-plugin/blob/08bfa36f6ed9f3838fcab484802df68f94fd0157/manifests/cmp-sidecar/cmp-plugin.yaml) it might be a good idea to append `-maxdepth 1` to the discover commands to prevent subdirectories from falsely triggering the plugin.

Also, for the helm plugin, it might not be necessary to require the `values.yaml` file, since some users may want to just use Chart.yaml for dependencies.

Issue #363, opened 2022-07-02
Generate on folder with empty Yaml


AlexanderProschek:
**Describe the bug**
When pointing Argo Vault Plugin at a directory with one or more exclusively empty yaml files, AVP throws an error and causes ArgoCD apps to enter into the `unknown` state. Having a source folder with exclusively empty yamls is not an issue with either a pure ArgoCD application, nor when manually deploying the manifests with kubectl, leading me to believe AVP should not have an issue with this either.

```
/tmp/argocd-vault-plugin on main !1 ❯ ls /tmp/example      
a.yaml
/tmp/argocd-vault-plugin on main !1 ❯ cat /tmp/example/a.yaml                                                    
/tmp/argocd-vault-plugin on main !1 ❯ go run main.go generate /tmp/example/                                                                                                              took 5s
Error: No manifests
Usage:
  argocd-vault-plugin generate <path> [flags]

Flags:
  -c, --config-path string         path to a file containing Vault configuration (YAML, JSON, envfile) to use
  -h, --help                       help for generate
  -s, --secret-name string         name of a Kubernetes Secret in the argocd namespace containing Vault configuration data in the argocd namespace of your ArgoCD host (Only available when used in ArgoCD). The namespace can be overridden by using the format <namespace>:<name>
      --verbose-sensitive-output   enable verbose mode for detailed info to help with debugging. Includes sensitive data (credentials), logged to stderr

exit status 1
/tmp/argocd-vault-plugin on main ❯
```

While `kubectl` is fine processing this file (doing nothing)

```
/tmp/argocd-vault-plugin on main ❯ kubectl diff -f /tmp/example
/tmp/argocd-vault-plugin on main ❯
```

As per [section 7.2 of the yaml 1.2 specification](https://yaml.org/spec/1.2.2/#72-empty-nodes):
> YAML allows the node content to be omitted in many cases. Nodes with empty content are interpreted as if they were plain scalars with an empty value. Such nodes are commonly resolved to a “null” value.

Looking into this, the root cause of the issue seems to be [these lines](https://github.com/argoproj-labs/argocd-vault-plugin/blob/main/cmd/generate.go#L78-L80=), so simply removing these lines would put AVP more in line with the behavior of kubectl.

**To Reproduce**
1. Create a folder with an empty yaml
2. Run `argocd-vault-plugin generate /tmp/example/`

**Expected behavior**
[...]

PR #350, opened 2022-06-02
feat: Verbose logging with flag

jkayani:
### Description
Adds a verbose output flag so that users can troubleshoot AVP usage easier. Goal is to be able to trace:

- The final configuration Viper sees
- The processing being done for each placeholder
- API calls and responses from backends for both authentication and getting secrets

Assuming this looks worthwhile, I think it'd be good for bug reports to include this info to help us debug what's happening. This PR amends the issue template to say that.

**Fixes:** #307 

### Checklist
Please make sure that your PR fulfills the following requirements:
- [ ] Reviewed the guidelines for contributing to this repository
- [ ] The commit message follows the [Conventional Commits Guidelines](https://www.conventionalcommits.org/en/v1.0.0/#summary).
- [ ] Tests for the changes have been updated
- [ ] Are you adding dependencies? If so, please run `go mod tidy -compat=1.17` to ensure only the minimum is pulled in.
- [ ] Docs have been added / updated
- [ ] Optional. My organization is added to USERS.md.

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] New tests
- [ ] Build/CI related changes
- [ ] Documentation content changes
- [ ] Other (please describe)

PR #364, opened 2022-07-02
Empty manifests should result in empty output

AlexanderProschek:
### Description
Removes the if statement to check if manifests are empty and updates the test accordingly

**Fixes:**
#363

### Checklist
- [x] Are you adding dependencies? If so, please run `go mod tidy -compat=1.17` to ensure only the minimum is pulled in.
- [x] Optional. My organization is added to USERS.md.

### Type of Change

- [x] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -75,10 +75,6 @@ func NewGenerateCommand() *cobra.Command {
 				return err
 			}
 
-			if len(manifests) == 0 {
-				return fmt.Errorf("No manifests")
-			}
-
 			for _, manifest := range manifests {
 
 				template, err := kube.NewTemplate(manifest, cmdConfig.Backend)
--- a/cmd/generate_test.go
+++ b/cmd/generate_test.go
@@ -80,7 +80,7 @@ func TestMain(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		expected := "No manifests"
+		expected := ""
 		if !strings.Contains(string(out), expected) {
 			t.Fatalf("expected to contain: %s but got %s", expected, out)
 		}
--- a/docs/howitworks.md
+++ b/docs/howitworks.md
@@ -337,3 +337,23 @@ spec:
       annotations:
         checksum/secret: <path:secrets/data/db#certs | sha256sum>
 ```
+
+### Error Handling
+
+#### Detecting errors in chained commands
+
+By default argocd-vault-plugin will read valid kubernetes YAMLs and replace variables with values from Vault.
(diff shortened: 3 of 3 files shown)


werne2j:
@AlexanderProschek Seems like a valid use case. Thanks for the contribution!

jkayani (commented):
The "No manifests" error was introduced here: https://github.com/argoproj-labs/argocd-vault-plugin/pull/231. If we take this patch, doesn't that mean we'll have this issue again with the AVP integration with other tools, right?

AlexanderProschek:
@jkayani if I understand correctly, what your change in #231 adds is the ability for AVP to act as a sort of guard against other programs mistakes and halt execution. By that same reasoning this should also fail:
```bash
~ ❯ ((>&2 echo "some error" && exit 1) | kubectl diff -f -); echo $?;
some error
0
```
which it does not. `kubectl` will happily take an empty yaml node.

My point (see #363) is that unlike AVP, other Kubernetes tools like `kubectl` and ArgoCD don't fail executing on an empty input, rather do exactly what the input tells them to do (empty in -> empty out). Convention requires errors to be handled before inputting things into a program.

May I suggest adding a debug flag that add some useful debugging info, including if the input is empty?

jkayani:
Re-reading your issue and re-visiting the problem from the previous PR, I think you make a good case. At that time I'd also forgotten about `set -o pipefail` which would've solved the problem without any AVP code changes - I ran this test to confirm using the release of AVP before that previous PR was merged:

```
$ sh -c '((>&2 echo "some error" && exit 1) | avp_old generate -c ~/Documents/ops/argo/vault/ibm-sm-creds.yaml -); echo $?;'
some error
0

$ sh -c 'set -o pipefail; ((>&2 echo "some error" && exit 1) | avp_old generate -c ~/Documents/ops/argo/vault/ibm-sm-creds.yaml -); echo $?;'
some error
1

$ avp_old version
argocd-vault-plugin v1.5.0 (89e02217591d856e3902be247df36ec7d489270f) BuildDate: 2021-11-01T15:53:36Z
```

So given that `kubectl` and Argo CD are OK with empty manifests, you're right, I think we should be too. It'd also be good to remind users (like past me) about `-o pipefail` to help them debug situations via the docs. We do have the verbose mode introduced in #350 but it's no use unless it's explicitly enabled. If you wouldn't mind adding to the docs as part of this PR, I'm happy to accept this.

AlexanderProschek:
@jkayani I added some docs in the howitworks README. Let me know if this is good or if you were looking for something else

jkayani (approved):

Issue #369, opened 2022-07-07
[Feature request] Provide sub-manifests with resources for people who want to provide their own argocd


jskrzypek:
**Is your feature request related to a problem? Please describe.**
I want to consume AVP via the kustomize manifest but AVP's manifest already includes the argocd manifest. I am using a different version of argocd (v2.4.4 now, but imagine it were a fork of their manifest or something). The following kustomization does not work, because kustomize will complain that the avp generator tries to provide the same resource id:

```sh
➜ cat <<EOF > kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: argocd
resources:
  - https://github.com/argoproj/argo-cd//manifests/ha/cluster-install?ref=v2.4.4
  - https://github.com/argoproj-labs/argocd-vault-plugin//manifests/cmp-sidecar?ref=9c7288a5b2d395fea19c1100f2cd07b547cc1ee2
EOF

➜ kustomize build .
Error: accumulating resources: accumulation err='accumulating resources from 'https://github.com/argoproj-labs/argocd-vault-plugin//manifests/cmp-sidecar?ref=9c7288a5b2d395fea19c1100f2cd07b547cc1ee2': URL is a git repository': recursed merging from path '/private/var/folders/yn/3qwz5t5j2_x__rdy__mpkf580000gn/T/kustomize-791362366/manifests/cmp-sidecar': may not add resource with an already registered id: ServiceAccount.v1.[noGrp]/argocd-application-controller.[noNs]
```

**Describe the solution you'd like**
If you provide the AVP specific resources/patches in a sub-manifest, then I would be able to point to that instead of the manifest that explicitly includes a reference to argocd, allowing me to provide my own. 

I.e. a folder structure like this:
```tree
manifests
├── cmp-configmap
│   └── kustomization.yaml                  # 1)
├── cmp-sidecar
│   └── kustomization.yaml                  # 2)
└── plugin
    ├── configmap
    │   ├── argocd-cm.yaml
    │   ├── argocd-repo-server-deploy.yaml
    │   └── kustomization.yaml              # 3) 
    └── sidecar
        ├── argocd-repo-server.yaml
        ├── cmp-plugin.yaml
        └── kustomization.yaml              # 4)
```

And the `kustomization.yaml`s as labeled:

1) manifests/cmp-configmap/kustomization.yaml
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
[...]

PR #370, opened 2022-07-07
Restructure manifest for composability

jskrzypek:
- manifests: Restructure to allow BYO argocd
- manifests: Restore root manifest -> cmp-configmap

### Description
I've introduced 2 changes in separable commits, pasting the commit messages below.

These could be considered refactors but I think they're a bit more feature-y... YMMV.
Let me know If I should reword the commits.

> feat[manifests]: Allow BYO argocd
>
> Restructuring the manifests to allow consumers of AVP to bring their own argocd
> base by using the manifests under `manifests/plugin`.
>
> This allows consumers to integrate more easily with other CMPs or alternate
> versions of argo that the AVP maintainers haven't patched the manifests to
> include.
>
> Because the resources they patch are now not included in the kustomization bases
> in manifests/plugins, the standard `patchesStrategicMerge` strategy will fail.
> To fix this we switch to the `patches` interface and provide an explicit target
> for each patch.

And

> feat[manifests]: Restore root manifest
>
> PR #352 introduced the sidecar pattern, but also renamed the prior default
> manifest, which was available through a reference to
> https://github.com/argoproj-labs/argocd-vault-plugin//manifests
>
> To get the same app, consumers would need to change any exsting references to
> https://github.com/argoproj-labs/argocd-vault-plugin//manifests/cmp-configmap
>
> This commit introduces a simple proxy manifest that makes that change backwards
> compatible, meaning that now older references to
> https://github.com/argoproj-labs/argocd-vault-plugin//manifests
> will still resolve to the same app state as
> https://github.com/argoproj-labs/argocd-vault-plugin//manifests/cmp-configmap
> rather than just breaking.

**Fixes:** #369

### Checklist

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] New tests
- [ ] Other (please describe)


jskrzypek:
Ok It does work now.

There's also another branch to look at here: https://github.com/argoproj-labs/argocd-vault-plugin/compare/main...jskrzypek:argocd-vault-plugin:jskrzypek/plugin-components

Which uses Kustomize Components, which are supposed to be more composable, instead of the base pattern

jkayani (commented):
Use case makes sense to me and I'm fine with the approach as-is because it's clearly understandable. I'm not that worried about having broken backwards compatibility with the Kustomize app since I'd imagine most people are using a specific tag but I'm also not opposed to fixing it here as you've done. 

I think it'd be good to mention in the docs that these options are available (point to the `manifests/cmp-*/plugin` dir if you want to handle the Argo CD installation yourself, use our "opinionated" Kustomize app in `manifests/cmp-*` if you don't). Otherwise this PR is good with me!

jskrzypek:
Great, I'll add a note to the docs tomorrow

jskrzypek:
Hmmm @jkayani I was trying to consume this PR in our argocd kustomize app, and I'm finding that it actually may not work. 

I'm using this:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: argocd
resources:
  - https://github.com/argoproj/argo-cd//manifests/ha/cluster-install?ref=v2.4.4
  - https://github.com/jskrzypek/argocd-vault-plugin//manifests/plugin/configmap?ref=jskrzypek/plugin-submanifests
```
But when I look at the argocd-repo-server, I don't see the initContainer that should be added by the manifest. 😞 

OTOH the Components strategy in my other branch does actually work when I do it this way:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: argocd
resources:
  - https://github.com/argoproj/argo-cd//manifests/ha/cluster-install?ref=v2.4.4

components:
  - https://github.com/jskrzypek/argocd-vault-plugin//manifests/components/sidecar?ref=jskrzypek/plugin-components
```

Can I ask you to double-check me on this? If you find the same, I'll just switch over the PR over to the components approach.

jskrzypek:
Ok, yeah, after several tests, this just doesn't seem to work as intended. I'm going to close this PR and reopen based on the other branch.

jskrzypek:
Closing in favor of #372

PR #409, opened 2022-10-10
Document additional helm parameters

jkroepke:
### Description

Add `--kube-version $KUBE_VERSION --api-versions $KUBE_API_VERSIONS` to avoid the helm capabilities issue on helm template.

Reference:
* https://github.com/argoproj/argo-cd/issues/1251
* https://github.com/argoproj/argo-cd/pull/3243
* https://github.com/helm/helm/issues/3377
* https://github.com/argoproj/argo-cd/pull/3318

Public source: https://github.com/argoproj/argo-cd/issues/8249

### Checklist

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -50,7 +50,7 @@ configManagementPlugins: |
       args: ["helm dependency build"]
     generate:
       command: ["sh", "-c"]
-      args: ["helm template $ARGOCD_APP_NAME . --include-crds | argocd-vault-plugin generate -"]
+      args: ["helm template $ARGOCD_APP_NAME . --include-crds --kube-version $KUBE_VERSION --api-versions $KUBE_API_VERSIONS | argocd-vault-plugin generate -"]
 ```
 For sidecar configured plugins, add this to `cmp-plugin` ConfigMap, and then [add a sidecar to run it](../installation#initcontainer-and-configuration-via-sidecar):
 ```yaml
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -90,7 +90,7 @@ configManagementPlugins: |
       args: ["helm dependency build"]
     generate:
       command: ["sh", "-c"]
-      args: ["helm template $ARGOCD_APP_NAME ${helm_args} . --include-crds | argocd-vault-plugin generate -"]
+      args: ["helm template $ARGOCD_APP_NAME ${helm_args} . --include-crds --kube-version $KUBE_VERSION --api-versions $KUBE_API_VERSIONS | argocd-vault-plugin generate -"]
```
werne2j, line 93: I think instead of continuing to add variables here to this example, maybe we can link to the helm docs? That we we don't have to call out all args since its just a helm template command and can rely on the helm docs.  We can  suggest using these particular options for the reason described in the PR.

jkroepke: I disagree here. The argo-vault-plugin documentation should have some kind of 'works out of the box' experience. No one will go through the helm docs and the helm docs does not describe the problem. The options and parameters are manadory, otherwise the code from documentation will lead into unexpected behaivors. Thats the point, why the options should be added there.

From my point of view, the general issue is that argo-vault-plugin can not re-use the existing helm logic from ArgoCD. But the dicussion feels off-topic here.

werne2j: @jkroepke can you also add to the manifests then? That would be the place for 'works out of the box' experience.

werne2j: > From my point of view, the general issue is that argo-vault-plugin can not re-use the existing helm logic from ArgoCD. But the dicussion feels off-topic here.

That is a larger Argo CD issue that we cant solve

PR #513, opened 2023-05-22
refactor(ibmsm): Support v2 API and KV secret types

jkayani:
### Description

IBM Cloud SM API v1 is going out of support in a few months and clients must migrate to v2: https://cloud.ibm.com/apidocs/secrets-manager/secrets-manager-v1. This implements the migration for AVP. It also adds support for KV secret types along with logging to indicate when "new" types, that are added by IBM Cloud SM that AVP might not support, are found and skipped. Finally, test for data races since `ibmsm` uses goroutines to speed things up.

Notable changes in the v2 API/SDK:

- Secret type isn't a necessary parameter to API calls (ID uniquely identifies a secret regardless of group/type)
- All secret types are versioned now, so AVP should not ignore version for arbitrary and other secret types when specified in placeholder/annotation
- The IBM Cloud SM SDK no longer returns the generic `SecretResource` struct `SecretData` key that allowed AVP to access the metadata and payload for (almost) any secret type. Instead structs specific to each type are returned (example: https://github.com/IBM/secrets-manager-go-sdk/blob/v2.0.0/secretsmanagerv2/secrets_manager_v2.go#L7211). While they all implement a common interface, it has 0 useful methods. This means AVP must handle each possible type explicitly. For this reason, the SDK return values are wrapped in new types `IBMSecretData`, `IBMVersionedSecretData`, `IBMSecretMetadata`; each type implements a generic `GetMetadata` or `GetSecret` method, so that the type-by-type code can be in 1 place and not each function
- Due to the above, only select keys (whatever was previously included in the `SecretData` in the v1 API) for each secret type are available for replacing a placeholder. This isn't configurable outside the codebase

### Checklist
- [x] Are you adding dependencies? If so, please run `go mod tidy -compat=1.17` to ensure only the minimum is pulled in.

### Type of Change

- [ ] Bugfix
- [x] Feature
- [ ] New tests
- [ ] Other (please describe)


werne2j (approved):
Looks good

PR #544, opened 2023-08-29
fix: look for vault cached token before login

DaThumpingRabbit:
### Description
This commit adds the check of the cached vault token before doing the authentication flow over again

**Fixes:** <https://github.com/argoproj-labs/argocd-vault-plugin/issues/536>

### Checklist

### Type of Change

- [x] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)

### DIFF
--- a/pkg/auth/vault/approle.go
+++ b/pkg/auth/vault/approle.go
@@ -33,7 +33,7 @@ func NewAppRoleAuth(roleID, secretID, mountPath string) *AppRoleAuth {
 
 // Authenticate authenticates with Vault using App Role and returns a token
 func (a *AppRoleAuth) Authenticate(vaultClient *api.Client) error {
-	err := utils.CheckExistingToken(vaultClient)
+	err := utils.LoginWithCachedToken(vaultClient)
 	if err != nil {
 		utils.VerboseToStdErr("Hashicorp Vault cannot retrieve cached token: %v. Generating a new one", err)
 	} else {
--- a/pkg/auth/vault/approle.go
+++ b/pkg/auth/vault/approle.go
@@ -33,6 +33,13 @@ func NewAppRoleAuth(roleID, secretID, mountPath string) *AppRoleAuth {
 
 // Authenticate authenticates with Vault using App Role and returns a token
 func (a *AppRoleAuth) Authenticate(vaultClient *api.Client) error {
+	err := utils.CheckExistingToken(vaultClient)
+	if err != nil {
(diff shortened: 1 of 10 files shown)


DaThumpingRabbit:
@werne2j @jkayani
Can you please run the workflow again ?
I have added tests on the use of cached tokens
Thanks !

DaThumpingRabbit:
Hello @werne2j @jkayani 
Any news on this ? Can you please run the workflow ?
Thank you 😄

werne2j:
@DaThumpingRabbit can you rebase your branch with main? Will run once up to date

DaThumpingRabbit:
Ah yes sorry I didn't notice the updates on the main branch, I just rebased it

DaThumpingRabbit:
Will you have a chance to review that ? I am waiting for this to reduce my vault lease consumption

werne2j:
@DaThumpingRabbit thanks for the PR. Looking it over, I think the approach I would like to see taken is to move all the token stuff (checking for and setting token) from the individual vault auth files to the vault backend https://github.com/argoproj-labs/argocd-vault-plugin/blob/main/pkg/backends/vault.go#L32. 

We then call the `v.AuthType.Authenticate` func if can’t find a token, we can change that func to return the token and then call the set token func with that token.

DaThumpingRabbit:
I see, I started to do those changes, however, I think that the token storing part was delegated to the individual auth files because of the specific token use-case (where the vault client detects the variable itself)
Do you think I should still extract this part to the vault backend file and make an extra case for the token to be handled differently ?

DaThumpingRabbit:
Or if this seems cleaner to you, we could add a method to the AuthType type to indicate whether the storing / retrieving of the token should be considered (maybe a ShouldStoreToken function that returns a boolean and the vault backend conditions its behavior according to that)

DaThumpingRabbit:
@werne2j I just committed a proposition to remove duplicated code in the vault auth types as you mentioned by following the idea I posted above
Let me know if this isn't ok for you and I can revert this commit

werne2j:
@DaThumpingRabbit Looking back over it, i think going with the first strategy is fine. No need to over think it. If you want to roll back to that commit we can kick off the CI and try to get this into the next release

DaThumpingRabbit:
@werne2j Thanks for checking it, I just rolled back to the first solution as you said and I am ready on my side if you want to trigger the CI again and merge it !

werne2j (approved):
Thanks @DaThumpingRabbit

PR #597, opened 2024-01-12
docs: use pipefail option in shell commands to prevent accidental deletions

jacksgt:
### Description

When using a pipe in shell commands (`|`) it is *critical* that the [pipefail](https://sipb.mit.edu/doc/safe-shell/) option is used, otherwise the second command in the pipeline silently overwrites any errors from the first command.
Argo CD will then treat the `stdout` output as valid (even if it is empty), which can have catastrophic effects: deleting all resources belonging to the Application/ApplicationSet.

Several users have already been hit by this:  https://github.com/argoproj-labs/argocd-vault-plugin/issues/533

Therefore I believe it is very important that the *correct* commands are shown in the documentation.

**Fixes:** https://github.com/argoproj-labs/argocd-vault-plugin/issues/533

### Checklist
- [ ] Are you adding dependencies? If so, please run `go mod tidy -compat=1.21` to ensure only the minimum is pulled in.

### Type of Change

- [ ] Bugfix
- [ ] Feature
- [ ] New tests
- [ ] Other (please describe)


docs/usage.md
```diff
@@ -34,9 +34,12 @@ spec:
   project: default
 ```
 
-3. Or you can pass the config-management-plugin flag to the Argo CD CLI app create command:  
+3. Or you can pass the config-management-plugin flag to the Argo CD CLI app create command:
 `argocd app create you-app-name --config-management-plugin argocd-vault-plugin`
 
+!!! warning
```
werne2j, line 40: Let’s use the bold formatting here for “warning” like we did with “important” below

jacksgt: Sure thing, done.

docs/usage.md
```diff
@@ -71,6 +73,8 @@ For sidecar configured plugins, add this to `cmp-plugin` ConfigMap, and then [ad
       generate:
         command:
           - sh
+          - "-o"
```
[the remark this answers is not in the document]
jacksgt: > Are you sure that 'pipefail' work with sh ?

Hi, thanks for spotting this!
I looked into the docs of  `bash` and `sh` (from alpine) and both of them support the `pipefail` option.
But indeed `dash` (which is the default `sh` on ubuntu/debian, which is used for Argo base images) does *not* support `pipefail`.

I have updated the PR to use `bash` everywhere where the `pipefail` option is used.

yyvess (approved):

PR #609, opened 2024-02-12
feat[ibmsm]: Secret group name resolution and simpler key interpolation

jgawor:
### Description

* Add support for using secret group names (instead of group IDs) as part of a path. For example, instead of `<path:ibmcloud/arbitrary/secrets/groups/123>` one can use `<path:ibmcloud/arbitrary/secrets/groups/myGroupName>`
* Add support for inserting values of non-arbitrary secrets without using the `jsonPath` modifier. For example, instead of `<path:ibmcloud/imported_cert/secrets/groups/123#my-cert-secret#previous | jsonPath {.certificate}>` one can use `<path:ibmcloud/imported_cert/secrets/groups/123/my-cert-secret/certificate#previous>`

The existing behavior  is unchanged.

### Checklist
- [ ] Are you adding dependencies? If so, please run `go mod tidy -compat=1.21` to ensure only the minimum is pulled in.

### Type of Change

- [ ] Bugfix
- [x] Feature
- [x] New tests
- [ ] Other (please describe)

### DIFF
--- a/pkg/backends/ibmsecretsmanager.go
+++ b/pkg/backends/ibmsecretsmanager.go
@@ -11,7 +11,8 @@ import (
 	"github.com/argoproj-labs/argocd-vault-plugin/pkg/utils"
 )
 
-var IBMPath, _ = regexp.Compile(`ibmcloud/(?P<type>.+)/secrets/groups/(?P<groupId>.+)`)
+var IBMPath, _ = regexp.Compile(`ibmcloud/(?P<type>.+)/secrets/groups/(?P<groupId>[^/\n]+)(/(?P<secretName>.+))?`)
+var GroupId, _ = regexp.Compile(`[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`)
 
 // IBMSecretMetadata wraps the SecretMetadataIntf provided by the SDK
 // It provides a generic method for accessing the metadata regardless of secret type
@@ -282,6 +283,7 @@ type IBMSecretsManagerClient interface {
 	ListSecrets(listAllSecretsOptions *ibmsm.ListSecretsOptions) (result *ibmsm.SecretMetadataPaginatedCollection, response *core.DetailedResponse, err error)
 	GetSecret(getSecretOptions *ibmsm.GetSecretOptions) (result ibmsm.SecretIntf, response *core.DetailedResponse, err error)
 	GetSecretVersion(getSecretOptions *ibmsm.GetSecretVersionOptions) (result ibmsm.SecretVersionIntf, response *core.DetailedResponse, err error)
+	ListSecretGroups(listSecretGroupsOptions *ibmsm.ListSecretGroupsOptions) (result *ibmsm.SecretGroupCollection, response *core.DetailedResponse, err error)
 }
 
 // Used as the key into the several caches for IBM SM API calls
@@ -314,6 +316,8 @@ type IBMSecretsManager struct {
 	// Keeps track of whether GetSecrets has been called for a given group and secret type
 	// Only read/written to by the main goroutine, no synchronized access needed
 	retrievedAllSecrets map[cacheKey]bool
+
+	secretGroups map[string]string
 }
 
 // NewIBMSecretsManagerBackend initializes a new IBM Secret Manager backend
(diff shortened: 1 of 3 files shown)


jkayani:
Thanks for the PR! I'll try to review within the week. Initial thoughts:

- Being able to use the group name instead of having to use the ID is a great addition - that way it's consistent with how secret interpolation works, where using the secret name can be used instead of the ID

- Being able to interpolate fields of the secret without using `jsonPath` is also nice. It might be a bit confusing that the IBM Cloud SM backend has 2 different path syntax options but I guess as long as the docs are clear it's OK. 

- On that note, please update the docs to describe the changes made here

Again, thanks for the contribution - code, tests and all. Glad to see this could be implemented without breaking existing behavior as well!

jgawor:
@jkayani I updated the docs. Please take a look.

pkg/backends/ibmsecretsmanager.go
```diff
@@ -536,18 +541,64 @@ func storeSecret(secrets *map[string]interface{}, result map[string]interface{})
 	return nil
 }
 
+func (i *IBMSecretsManager) ResolveGroup(group string) (string, error) {
```
jkayani, line 544: Does this function need to be exported outside the package? If not, let's keep it private like the others here

pkg/backends/ibmsecretsmanager.go
```diff
@@ -536,18 +541,64 @@ func storeSecret(secrets *map[string]interface{}, result map[string]interface{})
...
+	if group == "default" || GroupId.MatchString(group) {
+		return group, nil
+	}
+
+	// list groups
+	if len(i.secretGroups) == 0 {
+		opts := &ibmsm.ListSecretGroupsOptions{}
+		secretGroupCollection, _, err := i.Client.ListSecretGroups(opts)
+		if err != nil {
+			return "", fmt.Errorf("Could not list security groups: %s", err)
```
jkayani, line 555:
```suggestion
                        return "", fmt.Errorf("Could not list secret groups: %s", err)
```

pkg/backends/ibmsecretsmanager.go
```diff
@@ -536,18 +541,64 @@ func storeSecret(secrets *map[string]interface{}, result map[string]interface{})
...
+		}
+		for _, group := range secretGroupCollection.SecretGroups {
+			i.secretGroups[*group.Name] = *group.ID
+		}
+	}
+
+	// look up group id for group name
+	groupId := i.secretGroups[group]
+	if groupId == "" {
+		return "", fmt.Errorf("No such security group %s", group)
```
jkayani, line 565:
```suggestion
                return "", fmt.Errorf("No such secret group %s", group)
```

jkayani (commented):
Only minor feedback here - code, tests, and docs all look great. Once feedback's addressed this can be merged

I don't see any GH issue about this - what motivated this change? Does the alt path syntax you're adding here help in certain automation scenarios, etc? Just curious

Thanks!

pkg/backends/ibmsecretsmanager.go
jgawor: fixed.

pkg/backends/ibmsecretsmanager.go
jgawor: fixed.

pkg/backends/ibmsecretsmanager.go
jgawor: fixed.

jgawor:
@jkayani I addressed your comments and updated the PR. Thanks! As to the motivation for the PR, it was the overall ease of use. For group names, that's simple. Using the group ids is non-portable from one instance of IBM Secrets Manager to another (e.g. in different regions). Also, trying to decipher the security group id from a path (and mapping it to a group in IBM Secrets Manager) is much harder than a name. As to the simplified path syntax, we actually use the vault plugin with two backends: IBM and AWS. Injecting secrets with the AWS Secrets Manager is simple and does not usually `jsonPath` or other modifiers. With the existing IBM backend, the `jsonPath` modifier is pretty much always needed. So, the idea was to remove that requirement and maybe make moving between the backends a bit simpler.

jkayani (approved):
LGTM, thanks for your patience!

Paging @werne2j for his sign off

werne2j (approved):

